PR #97, opened 2016-12-12
Modified create issue view to handle anon users

rafen:
Now when an anon user submit a issue, we store it and we assign the issue to the user when the user logs in. To do that we use the user session.


seanauriti:
Thank you.  There is one issue, can you please fix this and then I will send the payment through Coderbounty - https://github.com/Bugheist/website/issues/99

rafen:
@seanauriti sure, I will fix that

rafen:
@seanauriti please see pull request https://github.com/Bugheist/website/pull/101

PR #101, opened 2016-12-13
Fixed visibility of incomplete bugs, ie: bugs with no associated user

rafen:
Since we are storing the bug of anon user and then we assign the user to the bug when he logs in. We might have some cases when user do not complete the loggin process, that generates bugs with no users, we can call them "incomplete" bugs.
I added two methods to the manager of the Issue Class: incomplete and complete to filter issues by state.
These two methos now are used on the views to filter the bugs and only show the complete ones.

### DIFF
--- a/website/models.py
+++ b/website/models.py
@@ -17,6 +17,7 @@ from django.core.files.base import ContentFile
 import requests
 from PIL import Image
 
+
 class Domain(models.Model):
     name = models.CharField(max_length=255, unique=True)
     url = models.URLField()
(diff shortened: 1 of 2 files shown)


seanauriti:
This is not the issue.  Coincidentally, we do want to allow for anonymous bug reporting so it's ok if they show up without a user attached, we can say "Anonymous" if there is no user.  The issue in this case was that I was logged out, submitted a bug and then logged in.  The bug was created correctly, however I was not associated with the bug as expected.

PR #104, opened 2016-12-28
Update http://www.coderbounty.com/issue/76/

ghost:
Hello, 

I have update this issue: http://www.coderbounty.com/issue/76/ . 
Please let me know if all ok or need changes.

Best regards,
Abel


abelradac:
FYI @seanauriti @mareiska

seanauriti:
Thank you! Can you please add the form to the bottom of the homepage?

seanauriti:
I've merged and sent the payment.  Thanks!

abelradac:
@seanauriti Can you please tell me how should it look at the bottom of page?
Example: Create a new panel like "Latest" and add form in it under "View all activity button" 

If example above is correct let me know and create pull request. 

Also i see a problem here. If i deleted invited emails every time 2 emails are invited how i know i will not reinvite the friend again and give points for nothing?

seanauriti:
Yes to first part. Second, I think the points should only be awarded when
the user signs up. So we can check if they are signed up already and change
the logic for the invite points. Thank you.

PR #149, opened 2017-02-25
Changed to View Website

rtgdk:
Changed to "view website"
#143


rtgdk:
Making both width and height of the screen shot 90% to fit better on any device size.
#147

seanauriti:
Thank you!

PR #157, opened 2017-05-28
Fix #125: Updated django-allauth to latest version.

### LINKED ISSUE #125: There is a validation during resetting the password that: The password so too similar to username but there is no any validation on setting the password fist time during sign up.
http://bugheist.com/issue/354

souravbadami:
The existing django-allauth (version: 0.27.0) was having an issue due to which `django.contrib.auth.password_validation.UserAttributeSimilarityValidator` was not working on registration. The latest version for django-allauth (version 0.32.0) fixes this problem. Fixes #125.


seanauriti:
Thank you!

PR #167, opened 2017-06-04
Updated tests.py for the new screenshot button.

mohitanand001:
I think the path for Imagepath in tests.py was configured on to work on Windows. I have changed the path to work on both Windows and Unix-like systems.


seanauriti:
Thanks!

PR #175, opened 2017-06-04
start button in about page made consistent with the design

Roboneet:
Hi , I've made the start button on about.html and hunt.html use the same design


souravbadami:
Hi @Roboneet -- From the design point of view, I think -- the button size is a bit big. Maybe it'll look better without the `btn-lg` class ? What's your opinion ?

Roboneet:
@souravbadami , Yeah , that would look better.

PR #179, opened 2017-06-04
Fixed Issue #121

mohitanand001:
Now the user cannot login without email confirmation.

### DIFF
--- a/bugheist/settings.py
+++ b/bugheist/settings.py
@@ -182,7 +182,7 @@ if DEBUG:
 
 ACCOUNT_EMAIL_REQUIRED = True
 ACCOUNT_USERNAME_REQUIRED = True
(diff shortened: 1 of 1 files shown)


souravbadami:
Hi @farziengineer -- I'm not sure if you got the issue #121. It says that, user can access the service entirely except he'll be shown a banner somewhere on top to confirm his email address.

farziengineer:
@souravbadami Hey Sourav, the issue heading says that "Without confirming the email address. user is able to login into the website", which should not happen. And yes even now they are being shown a banner to confirm their email address. I hope this is clear.

souravbadami:
If that's the case, then it seems legit. Not sure though. @seanauriti will give better insights on this.

Issue #186, opened 2017-06-05
Tests are broken


Roboneet:
As the user can't login without email verification(issue #121), the tests don't work now.

seanauriti:
I think it is better to allow the user to not have to verify email, and just show them a banner to verify it when they are on the site like twitter does.

farziengineer:
Which means the user can access the site without verifying email-id. Okay, then reverting the ACCOUNT_EMAIL_VERIFICATION to "OPTIONAL" would do the job.

souravbadami:
@farziengineer -- Indeed!

PR #189, opened 2017-06-05
Added a banner to remind users to verify email whenever they log in without verification

Roboneet:
as mentioned in #186


seanauriti:
Thank you!

PR #195, opened 2017-06-05
The issue #155 had popped up.

mohitanand001:


souravbadami:
Ah, Yes it has. Thanks! :)

PR #214, opened 2017-06-06
Refactor code in /index

jajodiaraghav:


souravbadami:
Hi @jajodiaraghav -- It looks good. How about moving the style configurations to `style.css` ? That way it will be more maintainable.

jajodiaraghav:
Will update the code in a jiffy!
Thanks @souravbadami

PR #222, opened 2017-06-07
Deleted the pyc files from comments app

mohitanand001:


seanauriti:
ImportError: No module named comments

farziengineer:
might be problem in settings.py. I'll check again.

farziengineer:
The problem is there because the __init__.py of the comments app is empty and hence can't be uploaded on github.

seanauriti:
Should be able too. Maybe add a #

farziengineer:
Okay

PR #252, opened 2017-06-08
updated readme

Roboneet:


souravbadami:
Hi @Roboneet -- Why exactly do we need to add `owasp.com` in domain for the dev server ?

Roboneet:
In index.html , I'm taking the number of open and closed requests in owasp through the context . So , the view uses 'Domain.objects.get(name="owasp.com")' . For that to work , we need owasp.com in  domain of dev server .

souravbadami (approved):
LGTM. Thanks!

PR #266, opened 2017-06-09
Fix #91: Add labels/badges on issues.

### LINKED ISSUE #91: Have badges for different types of bugs you've found
- See the error types on the homepage, there are 6

jajodiaraghav:
Fix #91


website/models.py
```diff
@@ -120,10 +120,20 @@ def validate_image(fieldfile_obj):
...
 
 class Issue(models.Model):
+    labels = (
+        (1, 'Number Error'),
+        (2, 'Functional'),
+        (3, 'Performance'),
+        (4, 'Security'),
+        (5, 'Typo'),
+        (6, 'Design'),
+        (0, 'Unspecified')
```
souravbadami: How about changing *Unspecified* to *General* ? I think, that will cover some of the bugs which doesn't comes under the predefined category.

website/templates/base.html
```diff
@@ -122,6 +122,15 @@
...
                     </div>
+                    <select name="label" class="form-control">
+                        <option value="0" selected="selected">Bug Type</option>
+                        <option value="1">Number error</option>
+                        <option value="2">Functional</option>
+                        <option value="3">Performance</option>
+                        <option value="4">Security</option>
+                        <option value="5">Typo</option>
+                        <option value="6">Design</option>
+                    </select>
```
souravbadami: This should be rendered from the backend instead of hard coding the properties here. It'll guarantee a safe submission of our attribute integrity on both the sides.

website/templates/issue_edit.html
```diff
@@ -64,6 +67,16 @@
...
+                <select name="label" class="form-control edit-label">
+                    <option value="0" selected="selected">Bug Type</option>
+                    <option value="1">Number error</option>
+                    <option value="2">Functional</option>
+                    <option value="3">Performance</option>
+                    <option value="4">Security</option>
+                    <option value="5">Typo</option>
+                    <option value="6">Design</option>
+                </select>
+
```
souravbadami: Ditto as the earlier comment.

website/forms.py
```diff
@@ -6,7 +6,7 @@ class IssueEditForm(forms.ModelForm):
 
     class Meta:
         model = Issue
-        fields = ['description', 'screenshot']
+        fields = ['description', 'screenshot', 'label']
```
souravbadami: It is a required element.

souravbadami (changes requested):
Hi @jajodiaraghav -- Looks great! I have left a few comments. Thanks!

website/forms.py
jajodiaraghav: I didn't get this part. Could be a bit more explicit?

souravbadami: We can set the attributes of the form in the backend. So, if we declare a *required* clause within the form definition in the backend -- It'll make sure that it is required when it is rendered and thus form submission is declined if it is not filled. It also solves the problem that you asked on *Slack*.
https://docs.djangoproject.com/en/1.11/ref/forms/fields/#django.forms.Field.required

website/templates/base.html
jajodiaraghav: Although I have rendered the form element in the edit page from the backend, I couldn't do it in base.html. I am new to Django and would love to see how to do this. Could you please change this part on top of my commit?

souravbadami: @jajodiaraghav -- What issue are you facing while using it in base.html ?

jajodiaraghav: {{form.label}} does not render anything in base.html. I did it in a way similar to the one in issue_edit.html. It took me a lot of time trying to debug, but all in vain. Can we merge this and fix the base.html in a different commit. I want to work on issue_edit.html page but couldn't, as it would result in conflicts unless this gets merged.

Or else, please help me fix this.

Thanks :)

souravbadami: Have a look at `website/forms.py` (IssueEditForm). The current implementation of the bug submission form needs a complete redesign. Maybe, we can do it completely on another PR.

souravbadami (approved):

PR #278, opened 2017-06-10
Removed tweets block.

souravbadami:


jajodiaraghav:
This looks good. Are you sure we should remove the tweets block?
@seanauriti

souravbadami:
@jajodiaraghav -- I had a discussion with **Sean** and we agreed on removing it from the home.

seanauriti:
I kind of like it because it shows that we tweeted it.  Anyone else have an opinion on it?

jajodiaraghav:
I think @souravbadami has a point. But if @seanauriti want to keep it, maybe we should place it somewhere else to make it a win win. Any suggestions?

souravbadami:
How about moving it to another page, called **social feeds**. Then, we can put it there in a column initially and then we can also add a Facebook feed by doing the same thing (posting issues on Facebook) ?

seanauriti:
Sounds good to me!

jajodiaraghav:
Sounds good!
Should this be merged then? We can create the social page in a different commit.

souravbadami:
Yes, let's merge this.

jajodiaraghav (approved):

PR #284, opened 2017-06-11
Fixed gravatar image on navbar.

souravbadami:


website/templates/base.html
```diff
@@ -56,9 +56,9 @@
                                 {% if request.user.userprofile.avatar %}
                                     <img src="{{ request.user.userprofile.avatar }}">
                                 {% elif request.user.socialaccount_set.all.0.get_avatar_url %}
-                                     <img src="{{request.user.socialaccount_set.all.0.get_avatar_url}}">
+                                    <img src="{{request.user.socialaccount_set.all.0.get_avatar_url}}">
                                 {% else %}
-                                    <img src="{% gravatar request.user.email 20 %}">
+                                    {% gravatar request.user.email 20 %}
```
jajodiaraghav: This is not the right fix. 
Please put this line: `` instead of `{% gravatar request.user.email 20 %}`

jajodiaraghav (changes requested):

jajodiaraghav:
@souravbadami Thanks for working on this. Left a comment :)

website/templates/base.html
souravbadami: This is the default `django_gravatar` approach which returns a pre-formatted tag for the element. See here: https://github.com/twaddington/django-gravatar.

jajodiaraghav: Sorry for not being explicit. What I meant was, we should change that to `` to maintain uniformity as this is the style we are following to spit out gravatar image within the app. Since no "class" or any other attribute is associated in this case, `gravatar request.user.email 20 %` works too.

Nevermind! 
Merging 🚀

jajodiaraghav (approved):

PR #286, opened 2017-06-11
Integrating Travis CI.

souravbadami:


souravbadami:
@jajodiaraghav -- This PR brings in a basic CI implementation and runs the simple test. I've commented out a test segment from the test for improving it as it is buggy in it's current state. I was wondering, have you tried to run it locally before ?

fredfalcon (approved):

seanauriti:
Thanks!

PR #338, opened 2017-07-09
Fix #336: Added new page for creating issues.

### LINKED ISSUE #336: Make a new page for creating issues.
Make sure it works with the chrome extension.
Also have an automated test to make sure it works.

souravbadami:


jajodiaraghav:
@seanauriti This looks good to me. Although, I still think the form on the header can be made cleaner. Or maybe we could reserve a section below the header for the form. Thoughts?

souravbadami:
Hi @jajodiaraghav -- Reserving more space below the header doesn't seems a nice idea to me. That place is already bloated. Maybe, we can make the current form more compact in some way.

fredfalcon (approved):

Issue #236, opened 2017-06-07
Remove inline CSS and JS to seperate files. Remove unnecessary static files


jajodiaraghav:
This will help developers' job easy.
Better code readability.

goyal-sidd:
I will try to ensure I work on this as I work about my proposal.

goyal-sidd:
Fixed in #348

PR #348, opened 2017-07-11
Removed inline CSS

sid22:
I have removed inline CSS from all the template html files. The some places left are dynamic. 
If I have missed anything kindly tell me.
Solves #236


jajodiaraghav:
> I have removed inline CSS from **all** the template html files

Are you sure there aren't anymore inline css present?
I bet there are hundreds of other places you missed. You could fix those in a different commit/PR if you want.

goyal-sidd:
Ok, I will check again in using Find.

goyal-sidd:
He @jajodiaraghav should I also work on inline style tags which have dynamic values. 
Like here in the screenshot:-
jajo

jajodiaraghav:
@goyal-sidd No. You do not have to remove them. Leave them as is

goyal-sidd:
I have removed most of style tags, the ones remaining are that are with dynamic entries or are single.

jajodiaraghav:
This looks much better. Thanks @goyal-sidd 
Yes, all inline css is removed :+1:

fredfalcon (approved):

PR #354, opened 2017-07-11
Added autocomplete using usernames.

### LINKED ISSUE #341: Autocomplete in comments section

mohitanand001:
screenshot from 2017-07-12 03 07 24
Fixes #341


farziengineer:
In the next patch, I'll be rendering the names as link to userprofiles like we have on github

jajodiaraghav:
@farziengineer Do you really think it is a good idea to send the entire user database as context data?
Currently we only have 600 users, but as the list of users increases, this page would take forever to load.
Rendering the list of 5 names using ajax is the best possible solution here.
You might wanna discuss this on chat.

farziengineer:
Sure I'm in college right now, I'll discuss asap.

fredfalcon (approved):

PR #361, opened 2017-07-13
Upgrade Django to 1.10

jajodiaraghav:


fredfalcon (approved):

seanauriti:
Can you please fix it so the tests work?

jajodiaraghav:
As explained on slack, actstream is not compatible with either Django 1.10 or 1.11. Hence the error:
```
File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/actstream/urls.py", line 5, in <module>
from django.conf.urls.defaults import url, patterns
ImportError: No module named defaults
```

We can easily fix the error by editing `/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/actstream/urls.py` and changing
this-> from django.conf.urls.defaults import url, patterns
to-> from django.conf.urls import *

I am not sure how to do this in travis. @souravbadami Can you help?

souravbadami:
@jajodiaraghav -- I think, it is not possible to do that within Travis or any other CI tool. Also, it is not a good practice too. Maybe, if it would have been possible to do that -- it'd be a bit overhead. Because, for each commit -- Travis builds the entire system for tests and installs the packages. So, changing that every time won't be easy (mind the easy keyword, because it can be easily done using a small script but not advisable).

Issue #145, opened 2017-02-22
. Make it so that before you submit a bug, if it is a company that has bugs already, you are taken to that page and all the bugs of that url are listed to check for duplicates


seanauriti:

farziengineer:
@seanauriti Can this be assigned to me.

souravbadami:
Hi @farziengineer -- Sure!

goyal-sidd:
Fixed in #362

PR #362, opened 2017-07-13
Functionality to see all bugs of the domain, if it exists

sid22:
Solves #145 . 
When the user goes to https://www.bugheist.com/report/ there a button is added below field for entering URL to check in db if bugs there for the domain entered.

Screenshots for way it works.

Case 1:
When bugs for the given domain exist:-
fe1
fe2

Case 2:
When bugs for the given domain don't exist:-
fe3

### DIFF
--- a/bugheist/urls.py
+++ b/bugheist/urls.py
@@ -5,7 +5,7 @@ from django.views.generic import TemplateView
 from website.views import (UserProfileDetailView, IssueCreate, UploadCreate, EmailDetailView, UpdateIssue,
                            InboundParseWebhookView, LeaderboardView, IssueView, IssueEditView, AllIssuesView,
                            HuntCreate, DomainDetailView, StatsDetailView, InviteCreate, CreateInviteFriend,
-                           ScoreboardView)
+                           ScoreboardView, domain_check)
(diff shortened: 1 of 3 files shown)


goyal-sidd:
Also @jajodiaraghav the form has CSRF however, while sending the POST request through AJAX to the function I have added a CSRF exempt. Is it alright? Since it's only a domain field.

jajodiaraghav:
@goyal-sidd No we do not need csrf for such ajax calls.
As mentioned earlier, you should never keep the code commented out in a commit, you must remove them before pushing since they are of no use.
Secondly, The styling of button and the message does not look right. You might want to fix those after this gets merged

goyal-sidd:
Removed comments, will ensure it doesn't get repeated. 
I will work on the styling on another issue.

souravbadami (approved):

jajodiaraghav:
This PR has some serious bugs. I don't understand why was it merged.

capture

@goyal-sidd Please fix this ASAP. Also, I can see some comments and console.log within the code, please remove them.

One more thing, you must learn the concepts of git `rebase` and `squash`. Please google these.

PR #395, opened 2017-07-20
Put footer to bottom

### LINKED ISSUE #391: Footer position
Footer bumps up on pages with less content than screen height.
footer_bump

jajodiaraghav:
Fixes #391

### DIFF
--- a/website/static/css/main.css
+++ b/website/static/css/main.css
@@ -1,3 +1,6 @@
+#page-wrapper {
+    height: 100%;
+}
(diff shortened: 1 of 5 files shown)


goyal-sidd:
Does it work for mobile view (height) etc. I was also working on it but was stuck for the mobile view only.

jajodiaraghav:
Yes it works for all views equally well

sid22 (approved):
Great!

PR #402, opened 2017-07-21
Types of bugs style fix

sid22:
newstyle

### DIFF
--- a/website/static/css/style.css
+++ b/website/static/css/style.css
@@ -513,3 +513,13 @@ input[type='file'] {
 .cardbox {
 	width: 20rem;
 }
+
+.cborder {
+    border-style: solid;
+    border-width: 1px;
+    padding-top: 15px;
+    border-radius: 5px;
+    margin-right: 25px;
+    margin-left: 25px;
+    margin-bottom: 10px;
+}
--- a/website/static/css/style.css
+++ b/website/static/css/style.css
@@ -513,13 +513,10 @@ input[type='file'] {
 .cardbox {
 	width: 20rem;
 }
-
 .cborder {
     border-style: solid;
     border-width: 1px;
     padding-top: 15px;
     border-radius: 5px;
-    margin-right: 25px;
-    margin-left: 25px;
-    margin-bottom: 10px;
+    margin-top: 25px;
 }
--- a/website/templates/index.html
+++ b/website/templates/index.html
@@ -22,11 +22,11 @@
         </div>
         <br>
     </section>
-
     <div class="container">
         <center>
             <div class="row">
-                 <div class="col-sm-3 cborder">    
+                 <div class="col-sm-4">
+                    <div class="cborder">    
                     <div class="card cardbox">
                       <center><i class="fa fa-exclamation-triangle fa-3x"></i></center>  
                       <div class="card-block">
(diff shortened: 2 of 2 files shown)


goyal-sidd:
Mobile view checked, button style consistent like rest of site.

website/templates/index.html
```diff
@@ -35,12 +35,10 @@ <h4 class="card-title">{% trans "Number Error" %}</h4>
...
                     </div>
                     <ul class="list-group list-group-flush">
-                        <li class="list-group-item">
-                            <a href="/search/?query=1&type=label">Examples</a>
-                        </li>
+                        <a class="btn btn-default" href="/search/?query=1&type=label">Examples</a>
                     </ul>   
                 </div>
-                <div class="col-sm-4">    
+                <div class="col-sm-3 cborder">
```
jajodiaraghav: Why are we changing from 4 to 3? Please read the bootstrap grids properly. Each row is divided into 12 sections (not 9), and hence the columns specified should add to 12.
Your changes will make the cards shift to left, leaving some space in the right making it inconsistent.

website/templates/index.html
```diff
@@ -26,7 +26,7 @@ <h1 class="page-header">{% trans "Find Bugs, Win Points and Prizes" %}</h1>
     <div class="container">
         <center>
             <div class="row">
-                <div class="col-sm-4">    
+                 <div class="col-sm-3 cborder">
```
goyal-sidd: We can target sm-4 but if we need it somewhere else without border then it won't be possible. 

We shouldn't change the bootstrap classes I think.

website/static/css/style.css
```diff
@@ -513,3 +513,13 @@ input[type='file'] {
 .cardbox {
 	width: 20rem;
 }
+
```
jajodiaraghav: Please remove this space. Have a look at the entire file, and then add the code in the similar format.
I am pretty sure that there is not new lines between different styles in the file.

website/templates/index.html
goyal-sidd: I have changed from 4 to 3 so there is some space between boxes without too much padding needed or risk of them overflowing. I have added margins to ensure they don't shift.

website/static/css/style.css
goyal-sidd: Ok. Will do it.

website/templates/index.html
jajodiaraghav: Just because we need the style somewhere else, that doesn't mean we have to add a new class for it. Almost every element in DOM can be targetted by the classes that are already there. Adding more classes is of no use.

goyal-sidd: Ok sir, I will look into it. and revert with the changes.

website/templates/index.html
jajodiaraghav: Don't you think that it looks bad with all the boxes shifted to the left with some unnecessary misaligned space at the right side?

jajodiaraghav: If we need padding in between, we can sure add padding just the way we add border.

goyal-sidd: I can add that way but in other screen sizes i thought it may overflow. I will try will sm-4 and check again with different method.

website/templates/index.html
jajodiaraghav: Just in case:
instead of making a new commit, use: `git commit --amend --no-edit`
And then push using `git push upstream branch_name -f`

goyal-sidd:
newchanged

jajodiaraghav:
@goyal-sidd Good work, this looks much better. Please push the code to let me read it

jajodiaraghav:
You didn't amend the commits as I asked you to!
I said instead of making a new commit, use `git commit --amend --no-edit`
And then push using `git push upstream branch_name -f`
:-1:

PR #424, opened 2017-07-23
Added cancel button when editing

mohitanand001:
Also the  edit and delete button will be hidden.

### DIFF
--- a/website/static/js/issue.js
+++ b/website/static/js/issue.js
@@ -77,6 +77,8 @@ $(function () {
         var old_message = $('#text_div_'+comment_id).text();
         $('#text_comment_'+comment_id).val(old_message);
         $('#text_div_'+comment_id).text('');
+        $(this).hide();
+        $(this).next('.del_comment').hide();
     });
 
     $(document).on('submit','.edit_form',function(f){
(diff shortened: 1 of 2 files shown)


website/static/js/issue.js
```diff
@@ -77,6 +77,8 @@ $(function () {
         var old_message = $('#text_div_'+comment_id).text();
         $('#text_comment_'+comment_id).val(old_message);
         $('#text_div_'+comment_id).text('');
+        $(this).hide();
+        $('#del_comment_'+comment_id).hide();
```
jajodiaraghav: This is the wrong way!
Remove the id from delete button.
Replace this `$('#del_comment_'+comment_id).hide();` with `$(this).closest('.del_comment').hide();`

jajodiaraghav (changes requested):

website/static/js/issue.js
farziengineer: If I go by your way, all the "delete" buttons will be hidden :(

jajodiaraghav: Okay, Just use: `$(this).next('.del_comment').hide();`

jajodiaraghav (approved):

PR #438, opened 2017-07-25
Add user-signup to stats

jajodiaraghav:
Also fixes 500 error


sid22 (approved):

sid22 (commented):
Sir, are you sure about the debug thing? Because, we  did that change when we got the error on main site.

fredfalcon (approved):

PR #523, opened 2017-09-03
Fix 500 Error in /profile

jajodiaraghav:


sid22 (approved):
lol, I was about to make a PR for same.

PR #543, opened 2017-09-05
mail fixes

sid22:
-added a more clear logo
-added description/label of bug content.
-made usernames/domains links to respective pages
-spacing between social links fixed
-more explanatory content in email body


fredfalcon (approved):

seanauriti:
Thanks, should we also separate back to txt and HTML  files ?

goyal-sidd:
We can but text will have lots of HTML elements for links etc, so separating now it won't be very useful. We should look to MailChimp templates eventually when we move to it. It would be lot more flexible approach.

PR #567, opened 2017-09-17
Reformatted codebase and optimised imports.

souravbadami:


fredfalcon (approved):
for the email script, because it was commented it removed the imports... nbd

Issue #447, opened 2017-07-27
ValueError: Missing staticfiles manifest entry for 'img/browser-logos/mobile safari/mobile safari_64x64.png'


fredfalcon:
View details in Rollbar: [https://rollbar.com/bugheist/Bugheist/items/70/](https://rollbar.com/bugheist/Bugheist/items/70/)

```
Traceback (most recent call last):
  File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/base.py", line 215, in _get_response
    response = response.render()
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/response.py", line 107, in render
    self.content = self.rendered_content
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/response.py", line 84, in rendered_content
    content = template.render(context, self._request)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/backends/django.py", line 66, in render
    return self.template.render(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py", line 207, in render
    return self._render(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py", line 199, in _render
    return self.nodelist.render(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py", line 990, in render
    bit = node.render_annotated(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py", line 957, in render_annotated
    return self.render(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/loader_tags.py", line 177, in render
    return compiled_parent._render(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py", line 199, in _render
    return self.nodelist.render(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py", line 990, in render
    bit = node.render_annotated(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py", line 957, in render_annotated
    return self.render(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/loader_tags.py", line 72, in render
    result = block.nodelist.render(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py", line 990, in render
    bit = node.render_annotated(context)
  File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py", line 957, in render_annotated
[...]

PR #572, opened 2017-10-21
Comment out whitenoise storage location in settings.py

rtgdk:
#447 
Since the image could not be found in the specified directory while calling
`python manage.py collectstatic`  , it throws an error. Commenting out the whitenoise line is a fix for it.


bugheist/settings.py
```diff
@@ -263,7 +263,7 @@
 
 # Simplified static file serving.
 # https://warehouse.python.org/project/whitenoise/
-STATICFILES_STORAGE = 'whitenoise.django.GzipManifestStaticFilesStorage'
+#STATICFILES_STORAGE = 'whitenoise.django.GzipManifestStaticFilesStorage'
```
souravbadami: @rtgdk - Is there any other way to fix it while keeping the whitenoise functionality ? Because it closes the entire whitenoise mechanism while solving issue for one image.

souravbadami (changes requested):

bugheist/settings.py
rtgdk: The fix is either to remove the line in the CSS which references the missing file, or to copy the file into the right place so it's no longer missing.
I'll look into it.

rtgdk: Or we can comment the `django.contrib.staticfiles` . We can also stop heroku from running collect static while pushing by  `heroku config:set DISABLE_COLLECTSTATIC=1`

PR #620, opened 2018-06-09
Removed slack key from travis configuration.

souravbadami:


fredfalcon (commented):
Why?

souravbadami:
Using this key, anyone can trigger notifications to the channel. Instead, we should encrypt the key. Documentation: https://docs.travis-ci.com/user/encryption-keys/.

PR #666, opened 2020-07-01
Develop -> Master

fredfalcon:


souravbadami (changes requested):
@fredfalcon - Can we merge this?

fredfalcon:
Please create a new homepage with the new design.  Let's separate out this branch to be only the new dashboard screens

PR #697, opened 2021-03-29
Trim search query

### LINKED ISSUE #689: Search query is not trimmed
The search query is not getting trimmed before displaying the results.
How to reproduce:
Case-1:
1. Search any random query in search box say: "amazon"
2. Note the search results.

image

Case-2:
1. Search the same query again but this time, add a whitespace before the query: " amazon"
2. Note the search results.

image

You will observe that the search results of Case-1 and Case-2 are not the same.
Expected Result: The results should be the same.

Sparsh1212:
Fixes #689: Trim search query

### DIFF
--- a/website/views.py
+++ b/website/views.py
@@ -873,6 +873,7 @@ class ScoreboardView(ListView):
 
 def search(request, template="search.html"):
 	query = request.GET.get('query')
+	query = query.strip()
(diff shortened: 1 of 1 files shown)


Sparsh1212:
@souravbadami @fredfalcon Please review.

PR #698, opened 2021-03-29
Add required attribute to Edit comment text area

### LINKED ISSUE #690: Disable Update button on empty string in Edit Comment
When we click on the Edit comment option, the Update button is enabled even on a blank string and does not display the "Please fill in this field" popup. 

image

Expected behavior: For an empty string case, it should be disabled and the message should be displayed while pressing the Update button, notifying the user that this field cannot be left blank.

Sparsh1212:
Fixes #690: Add required attribute to Edit comment text area

### DIFF
--- a/website/templates/comments.html
+++ b/website/templates/comments.html
@@ -16,7 +16,7 @@
                 <form class="edit_form" style="display:none">
                     {% csrf_token %}
                     <div class="form-group">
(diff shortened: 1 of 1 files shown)


Sparsh1212:
@souravbadami @fredfalcon Please review.

PR #772, opened 2021-06-13
Add Rest API Structure for User Authentication 

ankit2001:
This PR aims to add all basic API structure of registration, login , logout, user profile, and Social Apps.

Please review it @fredfalcon , @souravbadami .

Thanks


souravbadami (changes requested):
Lint tests are failing, could you please fix them?

ankit2001:
Yeah, I will fix them and add other APIs also.



[image: Mailtrack]
<https://mailtrack.io?utm_source=gmail&utm_medium=signature&utm_campaign=signaturevirality5&>
Sender
notified by
Mailtrack
<https://mailtrack.io?utm_source=gmail&utm_medium=signature&utm_campaign=signaturevirality5&>
16/06/21,
03:52:17 pm

bugheist/settings.py
```diff
@@ -272,7 +272,7 @@
 SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
 
 # Allow all host headers
-ALLOWED_HOSTS = [".bugheist.com", "127.0.0.1"]
+ALLOWED_HOSTS = ["*"]
```
fredfalcon, line 275: Please add this back, it makes it more secure

bugheist/settings.py
```diff
@@ -422,3 +433,12 @@
 STRIPE_LIVE_MODE = False  # Change to True in production
 
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
+
+CALLBACK_URL_FOR_GITHUB = os.environ.get(
+    "CALLBACK_URL_FOR_GITHUB", default="http://127.0.0.1:8000")
```
fredfalcon, line 438: What are these supposed to be?

fredfalcon (changes requested):

bugheist/settings.py
ankit2001: There are the callback urls for github and google, after authentication through google or github, they will redirect a user to some page with a code, we can use that code for access token, also can update that url to any thing where we want user to redirect after authentication.

bugheist/settings.py
ankit2001: I have done with this @fredfalcon .

bugheist/settings.py
fredfalcon: Can you default them to the base url?

ankit2001: Yes, we can remail them as it is.

ankit2001: Yes we can remail them localhost or default

ankit2001: Yes we can remain them localhost or default

fredfalcon:
Please fix the conflicts and it’s good to go, thanks!

fredfalcon (approved):

souravbadami (approved):

PR #795, opened 2021-10-19
remove unused variables and imports

spiderxm:


fredfalcon:
can you format it with black instead?

spiderxm:
sure

Issue #691, opened 2021-03-29
Upvoting issue does not increases its count without reloading


Sparsh1212:
While upvoting an issue, the count of upvotes is not increased unless the page is reloaded again.
Steps to Reproduce:
1. Go to any issue
2. Click the like button
3. We can see that the count is not increased
4. Reload the page
5. The count of likes gets updated
 
Expected behaviour: It should get increased without reloading

PR #890, opened 2022-06-12
Fix like count

AtmegaBuzz:
django html syntax wrong causing label = json object, which was throwing error and like button was not working.
 Upvoting issue does not increases its count without reloading #691


AtmegaBuzz:
#895

PR #892, opened 2022-06-12
changed issue view from templateview to IssueCreate

AtmegaBuzz:
report a bug page is missing captcha #738 
Before:
  when clicked on report bug page it was rendering html without added context["captcha"]
now:
   changed view from template to IssueCreate. now it is handled by  IssueCreate view which is adding context["captcha"]


AtmegaBuzz:
#895

Issue #891, opened 2022-06-12
Captcha Not Working


AtmegaBuzz:
How to reproduce this error:
1) go to https://www.bugheist.com/issue/
2) don't add any info just click submit.
3) captcha will appear with required field validations.
4) now add all the info and type anything gibberish in captcha.
5) now submit it. 
the form will submit.

PR #916, opened 2022-08-16
Fix/captcha validation

AtmegaBuzz:
As mentioned in #891  #738

Entered Invalid captcha "hello"
image

validation raised
image


fredfalcon:
Thank you!

Issue #930, opened 2022-08-22
make a leaderboard view that shows winner for each month for all time.


fredfalcon:
so like:
July 2022: winner: josh
June 2022: winner rogan
etc...
till the beginning of all data
it there was no winner say "no bugs found this month"

PR #932, opened 2022-08-23
each month leaderboard

AtmegaBuzz:
Issue:  make a leaderboard view that shows winner for each month for all time. #930 

image


fredfalcon (approved):
Nice work

PR #939, opened 2022-08-27
fixed global leaderboard

AtmegaBuzz:
prev it was filtering for current year in global board.
now its removed.


fredfalcon:
Thank you!

Issue #889, opened 2022-06-12
Wrong login credentials (Validation Message) not showing. 


AtmegaBuzz:
How to reproduce this error:
1) go to login 
2) enter wrong credentials 
3) no message displayed , screen refresh

PR #940, opened 2022-08-27
wrong credentials error message showing

AtmegaBuzz:
Issue :  Wrong login credentials (Validation Message) not showing. #889 

prev it was not showing error message
image

### DIFF
--- a/website/views.py
+++ b/website/views.py
@@ -555,8 +555,16 @@ class IssueCreate(IssueBaseCreate, CreateView):
             )
         return initial
 
-    def form_valid(self, form):
+    def post(self, request, *args, **kwargs):
         
+        if len(request.FILES['screenshot'].name)>99:
+            filename = request.FILES['screenshot'].name
+            extension = filename.split(".")[-1] 
+            request.FILES['screenshot'].name = filename[:88] + "." + extension
+
+        return super().post(request, *args, **kwargs)
+
+    def form_valid(self, form):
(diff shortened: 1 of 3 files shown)


fredfalcon:
I think the message is there we just need to show it.  We don’t need another view

AtmegaBuzz:
ok if thats the case then its not passing it in messages in the request. ill check

AtmegaBuzz:
@fredfalcon now error is showing without custom view 
image

website/views.py
```diff
@@ -20,6 +20,7 @@
 #from django_cron import CronJobBase, Schedule
 from allauth.account.models import EmailAddress
 from allauth.account.signals import user_logged_in
+from allauth.account.views import LoginView,_ajax_response
```
fredfalcon, line 23: Can remove this

fredfalcon (changes requested):
See comment

website/views.py
AtmegaBuzz: ok

AtmegaBuzz:
@fredfalcon resolved

fredfalcon (approved):

fredfalcon:
Thank you!

fredfalcon:
If you link the issue to the pr it will close automatically https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #943, opened 2022-08-31
large filename error fix

AtmegaBuzz:
Issue:  Automatically trim the filename for large file names so we don't get this error #741 
After: 
image
if name is >99 then it is trimmed downed to 88
image


fredfalcon (approved):

fredfalcon:
Thank you!

PR #983, opened 2022-11-26
Fix/tests

AtmegaBuzz:
- geckodriver removed from tests both tests now use chromedriver
- fixed all tests
- gitignore updated


fredfalcon:
Thank you!

PR #984, opened 2022-11-26
Run test on pr 

AtmegaBuzz:


website/views.py
```diff
@@ -558,7 +558,10 @@ def get_initial(self):
     def post(self, request, *args, **kwargs):
         
         # resolve domain
-        url = request.POST.get("url").replace("www.","https://")
+        url = request.POST.get("url")
```
fredfalcon, line 561: We need to remove www and http for all domains so they are stored the same in the db.

fredfalcon (changes requested):
Please change this back to how it was and only add the exception for bugheist

website/views.py
AtmegaBuzz: resolved

fredfalcon (approved):

fredfalcon:
Thank you!

PR #1012, opened 2023-01-03
Min-height to pages with less content

ManishBisht777:
Added min-height to the pages with less content

### changes

Added min-height to :

1) `signup.html`
2) `search.html`
3) `leaderboard.html`

I am using prettier to I ended up formatting the code as well :)

### DIFF
--- a/website/templates/account/signup.html
+++ b/website/templates/account/signup.html
@@ -1,88 +1,125 @@
-{% extends "base.html" %}
+{% extends "base.html" %} {% load i18n %} {% block style %}
+<style>
+  .container-min-height {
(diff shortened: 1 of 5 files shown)


fredfalcon (changes requested):
Can you just change the min height part and not the other code?

ManishBisht777:
actually, I am using a formatter so I did end up formatting the files. i will make the changes

ManishBisht777:
@fredfalcon hope it is good now :)

fredfalcon (approved):

fredfalcon:
Thank you!

PR #1097, opened 2023-03-02
Add url check api for checking for duplicate urls

### LINKED ISSUE #1096: API for domain_check
Need an API to check if a URL has already been reported, similar to this feature - 
image

letsintegreat:
Fixes #1096 

If the URL is found -

image

If the URL is not found -

image

### DIFF
--- a/blt/urls.py
+++ b/blt/urls.py
@@ -60,7 +60,8 @@ from website.api.views import (
     LikeIssueApiView,
     FlagIssueApiView,
     LeaderboardApiViewSet,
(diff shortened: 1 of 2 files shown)


AtmegaBuzz:
LGTM, i would suggest use better naming convention, don't use letters

letsintegreat:
> LGTM, i would suggest use better naming convention, don't use letters

Addressed, thanks

fredfalcon (approved):

fredfalcon (approved):

Issue #1131, opened 2023-03-13
issue when liking an issue


fredfalcon:
![image](https://user-images.githubusercontent.com/7475382/224851890-08328587-3047-425e-a18f-00041a98860c.png)

AtmegaBuzz:
Is it happening in mobile app? seems to be working on website

AtmegaBuzz:
@fredfalcon  error is due to no comma 
![image](https://user-images.githubusercontent.com/68425016/225083348-e8568509-aed9-4631-8b26-1277e57efc9b.png)

so the statement is evaluating to settings.EMAIL_TO_STRING[liked_user.email] and that's why getting index cannot be string

AtmegaBuzz:
its solved 
![image](https://user-images.githubusercontent.com/68425016/225961108-14cd07fb-747b-42a6-879e-eff2ca697f05.png)

PR #1132, opened 2023-03-14
Update views.py

AtmegaBuzz:
Fix: issue when liking an issue #1131


fredfalcon (approved):
good catch, thank you

PR #1213, opened 2023-04-10
Fix UI duplicated

AtmegaBuzz:
Added a pop up page which appear's when clicked on check for duplicate button. This popup takes the url entered and then query the backend API of UrlCheckApiViewset to get the list of top 10 latest issues . Future idea is to add a title in the Issue and then run a math algorithm to check for best match of (title,issuetype, domain) and return the latest results.

### DIFF
--- a/.github/workflows/assign-issues.yml
+++ b/.github/workflows/assign-issues.yml
@@ -3,6 +3,8 @@ name: Auto Assign Issues
 on:
   issue_comment:
     types: [created]
+  schedule:
+    - cron: '0 0 * * *'
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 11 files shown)


fredfalcon (approved):

AtmegaBuzz:
@letsintegreat The API format for check for duplicates is changed. Now it takes domain_url in post request and returns list of issue data 
image

letsintegreat:
> @letsintegreat The API format for check for duplicates is changed. Now it takes domain_url in post request and returns list of issue data 
> image
> 

Alright, I'll make the necessary changes in the app, thanks!

PR #1382, opened 2023-08-25
Fixing comment (everytime refresh to see the comment)

### LINKED ISSUE #739: Bug: When you comment, it does not show up initially, you have to refresh to see it.  It should show up right away
http://bugheist.com/issue/1523

Prakhar-Shankar:
Closes #739 

Now everytime you comment the window will refresh itself. Hence, the comment will be added without refreshing the page manually.

### DIFF
--- a/website/templates/issue.html
+++ b/website/templates/issue.html
@@ -46,6 +46,7 @@
         margin-left: 10px;
     }
 
+  
     .fa-white {
         color: white;
     }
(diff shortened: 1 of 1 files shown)


DonnieBLT:
Thanks for your contribution! The comment is submitted via ajax, so all we'd have to do is append the text to the end of the list.  Your solution still refreshes the page instead of appending it to the list without a refresh.  This task would be to append it to the list and not refresh the page.

Prakhar-Shankar:
> Thanks for your contribution! The comment is submitted via ajax, so all we'd have to do is append the text to the end of the list.  Your solution still refreshes the page instead of appending it to the list without a refresh.  This task would be to append it to the list and not refresh the page.  


Okay I got it, thanks for the information. I'll surely look into this.

PR #1396, opened 2023-08-30
made some changes in UI and improved visibility of page

officialarmannqureshi:
Before:
UI-02

After :
UI-02-Solution

2nd Improvement:
Before:

UI-03

After:

UI-03-Solution-Bg
UI-03-Solution-Border


AtmegaBuzz:
No need to do 2nd improvement, you can consult on slack about design. 1st change looks good.

Issue #1393, opened 2023-08-29
Bug: Endless/ Too long initial loading time


just-ary27:
As you might have seen upon opening the BLT website, the homepage takes forever to load. Not only is this inefficient but might cause problems to users opening website on a poorer connection.

**To Reproduce**
Just open the homepage of BLT website.

**Expected behavior**
Reduced loading time of homepage

CodeWithBishal:
Hi @just-ary27,
Just wondering if the issue still there?
I have checked and found the website loads fine for me: https://owasp.org/www-project-bug-logging-tool/

just-ary27:
That's the project information site, not the project site. We're talking about the project website in this issue

CodeWithBishal:
Hi @just-ary27, 

Just to confirm are we talking about www.bugheist.com here?

just-ary27:
Yes

CodeWithBishal:
@just-ary27  Can you assign this issue to me?

CodeWithBishal:
@officialarmannqureshi I am already working on this issue, was your plan to contribute to the code?

PR #1422, opened 2023-09-16
lazy loading images

CodeWithBishal:
Lazy loading images to boost performance. #1393 

We also need to enable severside gzip and that will boost the page performance significantly.


CodeWithBishal:
Please note: Enabling Server side gzip is a must to improve performance/loading speed significantly

DonnieBLT (approved):

DonnieBLT:
Thanks @CodeWithBishal !

PR #1436, opened 2023-09-25
Issue #828

JisanAR03:
BLT use for translate language :  i18n, i just make some changes on template i mean just rewrite some text with "{% trans '' %}" and also i did some changes on locale folder because i had translate some text manually because they are translate automatically .
I attach some image of before and after : 
    before: 
image
   after :
   
image

i fix all page like this for each language.


AtmegaBuzz:
Hey @JisanAR03 there are 300 + file changes, can you rebase to main

PR #1456, opened 2023-10-04
responsiveness error solved

CaduRoriz:
Tracking issue for: https://github.com/OWASP/BLT/issues/1451

Before: 
paginaquebrada

Now: 
paginaresolvida


AtmegaBuzz:
I think this will solve the current issue but there will change in behaviour for other elements using bootstrap, so would suggest if you can find a way to add this bootstrap in the language div only

AtmegaBuzz:
Wait there is already a PR raise on this Issue #1452

DonnieBLT:
closing as this was already fixed in another PR - reopen if I'm wrong

Issue #1469, opened 2023-10-11
Feature Request: Adding post coments deletion feature at issue2 page


JisanAR03:
**Is your feature request related to a problem? Please describe.**
at URL: <main_url>/issue2/<post_id> ,, comment deletion is not worked ,It should work with ajax call.

**Describe the solution you'd like**
Have to add a ajax call to backend and delete the comment by confirm it from a alert box.

**Additional context**
![image](https://github.com/OWASP/BLT/assets/97744811/b5fbc537-414a-4321-852d-c0dc557d0e99)

PR #1494, opened 2023-10-21
Adding post coments deletion feature at issue2 page

JisanAR03:
Fixed issue #1469 
Work:
    1.create a path for call the deletion request.(urls.py --- at website folder)
    2.create a function which return the comment data after delete the specific comment .(views.py --- at website folder)
    3.Write some jQuery code for make the ajax call(comments2.html --- at templates folder)
screenshot :
image

### DIFF
--- a/website/views.py
+++ b/website/views.py
@@ -2052,14 +2052,9 @@ def delete_comment(request):
     issue = Issue.objects.get(pk=int_issue_pk)
     if request.method == "POST":
         all_comment = Comment.objects.filter(issue=issue)
-        comment = Comment.objects.get(pk=int(request.POST['comment_pk']))
-        if request.user.username != comment.author:
-            return HttpResponse("You are not authorized to delete this comment")
-        try:
-            show = comment.parent.pk
-        except:
-            show = -1
-        comment.delete()
+        comment = Comment.objects.get(pk=int(request.POST['comment_pk']),author=request.user.username)
+        if request.user.username == comment.author:
+            comment.delete()        
     context = {
         "all_comment": Comment.objects.filter(issue__id=int_issue_pk).order_by("-created_date"),
         "object": issue,
--- a/website/views.py
+++ b/website/views.py
@@ -2050,11 +2050,10 @@ def update_comment(request, issue_pk, comment_pk):
 def delete_comment(request):
     int_issue_pk = int(request.POST['issue_pk'])
     issue = Issue.objects.get(pk=int_issue_pk)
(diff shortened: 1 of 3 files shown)


website/views.py
```diff
@@ -2047,6 +2047,24 @@ def update_comment(request, issue_pk, comment_pk):
     }
     return render(request, "comments2.html",context)
     
+def delete_comment(request):
+    int_issue_pk = int(request.POST['issue_pk'])
+    issue = Issue.objects.get(pk=int_issue_pk)
+    if request.method == "POST":
+        all_comment = Comment.objects.filter(issue=issue)
+        comment = Comment.objects.get(pk=int(request.POST['comment_pk']))
```
DonnieBLT, line 2055: Can you add the user to the query?

DonnieBLT (changes requested):

website/views.py
JisanAR03: > Can you add the user to the query?

yes sir sure.

website/views.py
```diff
@@ -2047,6 +2047,19 @@ def update_comment(request, issue_pk, comment_pk):
     }
     return render(request, "comments2.html",context)
     
+def delete_comment(request):
+    int_issue_pk = int(request.POST['issue_pk'])
+    issue = Issue.objects.get(pk=int_issue_pk)
+    if request.method == "POST":
+        all_comment = Comment.objects.filter(issue=issue)
+        comment = Comment.objects.get(pk=int(request.POST['comment_pk']),author=request.user.username)
+        if request.user.username == comment.author:
```
DonnieBLT, line 2056: Can remove this line as the query verifies it.

website/views.py
```diff
@@ -2047,6 +2047,19 @@ def update_comment(request, issue_pk, comment_pk):
     }
     return render(request, "comments2.html",context)
     
+def delete_comment(request):
+    int_issue_pk = int(request.POST['issue_pk'])
+    issue = Issue.objects.get(pk=int_issue_pk)
+    if request.method == "POST":
+        all_comment = Comment.objects.filter(issue=issue)
```
DonnieBLT, line 2054: All comment is loaded below so you can remove this too

DonnieBLT (changes requested):

website/views.py
JisanAR03: "all_comment" for sir fetch whole comment is web page after delete the comment

DonnieBLT (approved):

DonnieBLT (approved):

Issue #1553, opened 2023-11-01
Fix code scanning alert - URL redirection from remote source


DonnieBLT:
Tracking issue for:
- [ ] https://github.com/OWASP/BLT/security/code-scanning/88

![image](https://github.com/OWASP/BLT/assets/128622481/dd6f23b7-12e0-4440-a2c1-0caf216b6157)

caue96:
Hi @DonnieBLT. I just want to make sure that what I was working on in the past few days was correct. This issue, [#1330](https://github.com/OWASP/BLT/issues/1330) and [#1331](https://github.com/OWASP/BLT/issues/1331) just can be solved after implementing facebook, github and google signup and signin. Is that right?

JisanAR03:
> Hi @DonnieBLT. I just want to make sure that what I was working on in the past few days was correct. This issue, [#1330](https://github.com/OWASP/BLT/issues/1330) and [#1331](https://github.com/OWASP/BLT/issues/1331) just can be solved after implementing facebook, github and google signup and signin. Is that right?

if you complete "signup and signin" part ,, can you make a PR ?

DonnieBLT:
> Hi @DonnieBLT. I just want to make sure that what I was working on in the past few days was correct. This issue, [#1330](https://github.com/OWASP/BLT/issues/1330) and [#1331](https://github.com/OWASP/BLT/issues/1331) just can be solved after implementing facebook, github and google signup and signin. Is that right?

no, these are new features that require coding the feature

PR #1575, opened 2023-11-05
add https connection and allowed hosts check

MattSilverio:
Resolving problem to issue #1553


DonnieBLT (changes requested):
The security issue highlighted concerns untrusted URL redirection from a remote source, where user input is incorporated into a URL redirect without full validation, leading to potential phishing attacks (CWE-601). The proposed change ensures that the scheme is HTTPS and the netloc is in the allowed hosts. However, this may not fully mitigate the risk as it does not address the validation of the entire URL, such as the path and query parameters which can also be exploited.

To address the issue more comprehensively, consider implementing the following:

1. Validate the full URL structure, not just the scheme and netloc.
2. Ensure that the path and query parameters do not redirect to an untrusted page even if the netloc is within the allowed hosts.
3. Use a list of allowed paths or a regular expression to validate the entire redirection endpoint.
4. Employ a safe redirect function that only redirects to known-safe URLs defined within the application.

Implementing these additional checks will help to ensure that the application is not vulnerable to open redirection attacks stemming from user input.

JisanAR03:
@MattSilverio  the github , google and facebook callback function have to be updated. What you have to do is : read those social media documents and implement it on the BLT, which I'm going gonna do after few days, or if you can do that, implement this one, but don't do  something like sanitization system cause this is mainly feature problem, not a security one. best of luck ☺

MattSilverio (commented):
views.py with additional checks as requested

MattSilverio:
> The security issue highlighted concerns untrusted URL redirection from a remote source, where user input is incorporated into a URL redirect without full validation, leading to potential phishing attacks (CWE-601). The proposed change ensures that the scheme is HTTPS and the netloc is in the allowed hosts. However, this may not fully mitigate the risk as it does not address the validation of the entire URL, such as the path and query parameters which can also be exploited.
> 
> To address the issue more comprehensively, consider implementing the following:
> 
> 1. Validate the full URL structure, not just the scheme and netloc.
> 2. Ensure that the path and query parameters do not redirect to an untrusted page even if the netloc is within the allowed hosts.
> 3. Use a list of allowed paths or a regular expression to validate the entire redirection endpoint.
> 4. Employ a safe redirect function that only redirects to known-safe URLs defined within the application.
> 
> Implementing these additional checks will help to ensure that the application is not vulnerable to open redirection attacks stemming from user input.



> views.py with additional checks as requested

MattSilverio (commented):
Additional checks implemented as requested, @DonnieBLT

MattSilverio (commented):
Check last commits

.github/workflows/codeql.yml
DonnieBLT, line 1: we have a codeql file already I think you can remove this

DonnieBLT (commented):
looks good, just let me know about the file

fredfalcon (approved):

DonnieBLT (approved):

Issue #1314, opened 2023-08-12
Fix code scanning alert - Full server-side request forgery


DonnieBLT:
Tracking issue for:
- [ ] https://github.com/OWASP/BLT/security/code-scanning/40

![image](https://github.com/OWASP/BLT/assets/128622481/604acb1c-fbbe-442b-8164-13698af94c36)

JisanAR03:
I going to fix this one today

JisanAR03:
fixed by #1580

PR #1577, opened 2023-11-05
fixing server-side forgery

Prakhar-Shankar:
Closes issue #1314


JisanAR03:
@Prakhar-Shankar  sir , can you please check it by create a new issue cause your provided solution don't allow user to create issue which is very important for BLT

Prakhar-Shankar:
> @Prakhar-Shankar sir , can you please check it by creating a new issue, Your provided solution don't allow user to create issues, which is very important for BLT

It says domain dont exist, while reporting a new bug

Issue #1101, opened 2023-03-04
upvote and downvote on issues (show like counts, and add a dislike and show counts) (different than flag)


fredfalcon:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Additional context**

letsintegreat:
> do we already have this?

Yes, I believe "like" and "flag" on the issue page is equivalent to "upvote" and "downvote" respectively.

DonnieBLT:
dislike would be different than flag - so for this we can have like and dislike next to each other and add the dislike functionality.

PR #1598, opened 2023-11-17
Added dislike feature (for #1101)

SahilDhillon21:
I have added the dislike feature, and also made a small enhancement: initially, when a user clicks on the like button, we see the counter go up, but in the "Liked By" modal the user is not present and requires a reload to be visible. This now happens immediately after liking/disliking, without the need for reload.

Changes:

1. Added an "issue_downvoted" field in the UserProfile model
2. Added a dislike_issue url and corresponding view. It makes sure that if user has already liked the issue and then clicks dislike, the like gets removed and updated instantly. Same for the other way round.
3. Added the dislike html and combined it to form a single _likes_and_dislikes.html file.
4. In order to update both buttons in case a situation like that in point 3 arises, I have wrapped the buttons inside a span with class "like_dislike_buttons" and the latest context is passed in whenever a button is pressed to make sure the changes are instantly visible.
5. Also changed the "Flag" button logo to a flag, since we wanted that feature separately.

Screenshots:
[
o1
o2
](url)

A dislike doesn't notify the issue poster through email, but can be added easily if needed (user anyway doesn't get notified when someone flags the issue, so I figured even a dislike doesn't need to be notified).


DonnieBLT (approved):

DonnieBLT:
Nice work! Thank you.

Issue #1617, opened 2023-11-29
New Home Page for BLT


errorassassin:
We have decided not to continue with the current home page and build a new one. This will open new sections to the website and give it a better look overall.

### New Home Page Task-List
- [x] Design - can be found on this [Figma Link](https://www.figma.com/file/JE0QlQ95FVpk0wG9HyFw4D/BLT-Website?type=design&node-id=0%3A1&t=9aPMxdjUDBIUirAz-1)
- [ ] Change File Structure for accommodating the new home page at www.bugheist.com/newhome/ until complete production is completed.
- [ ] Build the header.
- [ ] Build the sidebar.
- [ ] Build the layout of bugs (dummy data).
- [ ] Show fetched data in the layout.
- [ ] "Report an issue" box.
- [ ] Leaderboard.
- [ ] Design for mobile.
- [ ] Make the page responsive.
- [ ] Make this homepage the default landing page of BLT.

errorassassin:
assign to me

PR #1618, opened 2023-11-29
New Home Page file structure and Header

errorassassin:
Towards #1617
The new home page can now be accessed on www.bugheist.com/newhome/, with a new header (for both signed-in and out).


DonnieBLT:
The newer one has been approved and merged

PR #1625, opened 2023-12-03
Created the layout for bugs on the new home page.

errorassassin:
Towards #1617. 
The new home page is populated with dummy bugs which looks like this:
image


DonnieBLT (approved):
Nice work 🏆

PR #1634, opened 2023-12-06
fix new home page few things

JisanAR03:
work on new home page : 
1. by click on user name --redirect to user profile
2. click on post title user will redirect to new issue single page
3. twitter share button in new home page
4. now title will show 1st 100 latter 

screenshot: 
image


DonnieBLT (approved):
Thank you!

PR #1639, opened 2023-12-08
Fix like in new page

JisanAR03:
work : now user can like or upvote an issue from new home page

screenshot: 
image


JisanAR03:
@DonnieBLT  sir, can you please check this PR , I added "like" system to issue on the new home page

DonnieBLT (approved):
Thank you! And thanks for the nudge.

PR #1658, opened 2023-12-22
like and flag fix

errorassassin:
there was a typo in flagged (has been changed to issue_flaged intentionally).

Now likes and flags retain and update which was not the case before.
Towards #1617


website/alternative_views.py
```diff
@@ -66,7 +66,9 @@ def flag_issue2(request, issue_pk):
     issue_pk = int(issue_pk)
     issue = Issue.objects.get(pk=issue_pk)
     userprof = UserProfile.objects.get(user=request.user)
+    print(UserProfile.objects.filter(issue_flaged=issue))
```
DonnieBLT, line 69:
```suggestion
```

website/alternative_views.py
```diff
@@ -66,7 +66,9 @@ def flag_issue2(request, issue_pk):
...
     if userprof in UserProfile.objects.filter(issue_flaged=issue):
+        print("already flagged")
```
DonnieBLT, line 71:
```suggestion
```

DonnieBLT (changes requested):

DonnieBLT (approved):

PR #1689, opened 2024-01-22
Fixed the auto assign GitHub action.

kr-2003:
@DonnieBLT 
I tried to fix the auto assign issues.
Automatic Issue Assign will work when the following conditions are met:
1 - If the issue has comment saying '/assign', 'assign to me', 'please assign this to me' etc., then the issue will be automatically assigned.
### OR
2 - If the issue already has a pull request, then it will be automatically assigned to the one who created the pull request.
Am I right?
Please review the changes made.


DonnieBLT:
That's not quite right. Currently if the issue is assigned and a PR is attached to a different, it won't let the user assign.  I think the change for this would be in the BLT-Action repo.

DonnieBLT (changes requested):
Check comment

DonnieBLT:
No. It would be if the person is assigned to an issue and that issue has a pr, it would let them assign it to themselves. If there is no pr, it operates the same, only allowing one issue to be assigned.

kr-2003:
@DonnieBLT 

For example, if User A is already assigned to an issue and User B creates a pull request (PR) attached to the same issue, the Auto Assign feature should allow User A to assign the issue to themselves. This applies specifically when there is a PR associated with the issue. On the other hand, if there is no PR, the behaviour remains unchanged, allowing only one person (in this case, User A) to be assigned to the issue.

Am I correct??

DonnieBLT:
We can focus on only the user A scenario. Please check the extension repo for the current update on this issue and apply the fix there. Thank you. Closing this out.

Issue #1606, opened 2023-11-24
Task: Update Django all auth


DonnieBLT:

ATHARVA-GAWAS:
Please elaborate about this issue. I am interested on working about this issue.

Deus1704:
@DonnieBLT sir this task can be marked completed/closed

PR #1697, opened 2024-01-26
updated and fixed django-allauth.

Deus1704:
This PR is for #1606 
I have updated the Django-allauth to 0.60.1 and the corresponding requirement dj-rest-auth to 5.0.2 and included the middleware necessary for the new allauth to function properly.


DonnieBLT (dismissed):

Deus1704:
@DonnieBLT Sir I think I get why the Django Test/build is failing. The issue is that the api endpoint /auth/registration is returning No Content code=204, but the console registers the user's credentials.
image

 Ideally the api should be returning the key. I am trying to understand the apis and will find a fix.

Currently:
image

Ideally:
image

DonnieBLT:
@Deus1704 we may have do update this at the same time - https://github.com/OWASP/BLT/pull/1594

Deus1704:
> @Deus1704 we may have to update this at the same time - #1594

Yeah, this would close both the PRs at the same time.

DonnieBLT (approved):
Thank you!

Issue #1685, opened 2024-01-20
make sure that privately reported issues do not show up in the public image folder


DonnieBLT:

HanilJain:
it would be really helpful if you could provide a link.

HanilJain:
fixed #1691

DonnieBLT:
I think the best way to do this would be to create a new private bucket

HanilJain:
what does exactly new private bucket means ?

DonnieBLT:
A private bucket, in the context of cloud computing and storage, typically refers to a storage container within a cloud storage service that is designed to hold data. Unlike public buckets, which can be accessed by anyone with the right URL, private buckets are restricted and can only be accessed by specific, authorized users or systems. 

Key features of a private bucket include:

1. **Access Control:** The owner of the bucket can set permissions to control who can view, upload, or download data from the bucket.

2. **Security:** Private buckets often have enhanced security measures like encryption, both at rest and in transit, to protect sensitive data.

3. **Data Integrity:** They often include features to ensure the integrity of the data stored, like versioning and checksums.

4. **Integration:** These buckets can be integrated with other cloud services for data processing, analysis, or backup.

5. **Cost:** The cost of using a private bucket can vary based on the amount of data stored, the level of access control, and additional security features.

Private buckets are commonly used by businesses and individuals to store sensitive data like personal information, confidential business documents, or proprietary data, ensuring that it's not publicly accessible or vulnerable to unauthorized access.

HanilJain:
@DonnieBLT I think this issue can be closed, since #1691 is merged

DonnieBLT:
No, this issue is still valid - we'll have a separate private bucket for private issues

PR #1729, opened 2024-01-30
On hidden issues, Request access feature 

HanilJain:
### Fixed #1090 

- on accessing hidden issues it will direct them to page(issue/<int:pk>/request_access/)
  with mentioning < issue has been hidden by the user that reported it.>

- on requesting access it will email to the user for the access


HanilJain:
## Additional Feature to be added 

which will be fixed in this issue #1685 

#### Feature description 

- individual user can see their hidden issues in dedicated page to it 
- user could edit access to issues and user you can view it

blt/urls.py
```diff
@@ -317,7 +318,7 @@
     path("issue/<str:issue_pk>/comment/update/<str:comment_pk>/", website.views.update_comment, name="update_comment"),
     # delete_comment 
     path("issue2/comment/delete/", website.views.delete_comment, name="delete_comment"),
-    re_path(r"^issue/(?P<slug>\w+)/$", IssueView.as_view(), name="issue_view"),
+    path("issue/<int:slug>/" , IssueView.as_view() , name = "issue_view"),
```
DonnieBLT, line 321: Can keep this how it is

DonnieBLT (changes requested):
This looks similar to the other whole request. Can you please close the one that is not required and only keep the one that has all the changes required open?

HanilJain:
closing this pull request as all additional feature including this have been added in #1737

PR #1747, opened 2024-02-05
Rate Limiting added for entire website

ATHARVA-GAWAS:


website/views.py
```diff
@@ -88,6 +88,13 @@
 from django.core.exceptions import ValidationError
 from django.http import HttpRequest
 
+from django.views.generic.edit import CreateView
+from django.http import HttpResponse
```
DonnieBLT, line 92: These seem like they can be removed

DonnieBLT (changes requested):
Tweaks

ATHARVA-GAWAS:
I have made the suggested changes

DonnieBLT (changes requested):
Please remove the remaining unused imports

ATHARVA-GAWAS:
I have made the changes

DonnieBLT (changes requested):
Why is this removing the stype code?

DonnieBLT (changes requested):
We want this to be site wide. Can you please add it to the middleware? Also please remove any white space changes (changes to lines with no code)

Issue #1239, opened 2023-05-06
post to twitter


DonnieBLT:

keshavagr273:
Hey @DonnieBLT could you please describe this issue?

DonnieBLT:
this used to work, not sure of the updated terms, but we would like an option when reporting a bug to [x] post to twitter and it will post to the website's twitter account.

HanilJain:
![Screenshot 2024-02-13 004600](https://github.com/OWASP/BLT/assets/119354421/e192ccfc-0095-46e6-bdb3-1c10d8d64614)

@DonnieBLT 
isn't there already an Tweet option here ?
what else do we need to change ?

DonnieBLT:
This is an auto tweet when the issue is created

HanilJain:
@DonnieBLT  Is it an auto tweet on BLT twitter page ?
where it mentions that an issue was reported on BLT and here the link for it .

PR #1772, opened 2024-02-14
Tweet Feature 

HanilJain:
### Added Feature #1239 

using **tweepy** library this feature has been added.

BEARER_TOKEN
APP_KEY
APP_KEY_SECRET
ACCESS_TOKEN
ACCESS_TOKEN_SECRET
needed to be set in the .env file

here is an example video 

https://github.com/OWASP/BLT/assets/119354421/6d943dc3-3693-4bea-9457-ba53a841c092


HanilJain:
@DonnieBLT could you please review

DonnieBLT:
Can you please add instructions to get those values?

DonnieBLT (changes requested):
I think we already had sent to Twitter somewhere else in the code. Can you check where that is and see if that's working and maybe updates at that section

HanilJain:
> I think we already had sent to Twitter somewhere else in the code. Can you check where that is and see if that's working and maybe updates at that section

yes there was but not was used anywhere.
and dose not seems to contain API v2 and complicated 
So I removed it.
is it okay ?

HanilJain:
> Can you please add instructions to get those values?

https://developer.twitter.com/en/docs/authentication/oauth-1-0a/api-key-and-secret

https://developer.twitter.com/en/docs/authentication/oauth-2-0/bearer-tokens

DonnieBLT (approved):

Issue #1708, opened 2024-01-27
Project: Maintenece


DonnieBLT:
This is for tasks / bugs that don't belong to a specific project.

#Tasks:

#Bugs

Prakhar-Shankar:
@DonnieBLT  Hi sir, can I work on unit testing? Or is it already done?

Issue #888, opened 2022-06-11
Task: implement rate limiting for posting issues and in general for the website / api


fredfalcon:

DonnieBLT:
This is to implement a rate limiting plugin.

JisanAR03:
@DonnieBLT sir, I just want to know that if I do something like a user or anonymous user can create 10 (as an example) issues per day through the mobile app (api) and BLT website, is this fulfilling this task?

PR #1787, opened 2024-02-17
Rate Limiting for entire website

ATHARVA-GAWAS:
Fixes issue

#1708 #888

### DIFF
--- a/website/tests.py
+++ b/website/tests.py
@@ -13,6 +13,10 @@ from selenium.webdriver.support.ui import WebDriverWait
 
 from .models import Issue, IssueScreenshot
 
+from website.views import IssueBaseCreate
+from rest_framework.test import APITestCase,APIRequestFactory
+from rest_framework import status
+
 os.environ['DJANGO_LIVE_TEST_SERVER_ADDRESS'] = 'localhost:8082'
 
 class MySeleniumTests(LiveServerTestCase):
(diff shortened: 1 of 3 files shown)


website/tests.py
```diff
@@ -131,3 +135,16 @@ def test_on_hide(self):
 
             if "hidden" not in filename:
                 self.assertFalse(True, "files rename failed")
+
+class IssueBaseCreateTest(APITestCase):
+    def test_throttle_exceed_limit(self):
+        for i in range(10000):
+            data = {'url': 'http://example.com', 'description': 'test', 'markdownInput': 'test', 'screenshots': 'test', 'captcha_1': 'PASSED'}
+            response = self.client.post('/report/', data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
DonnieBLT, line 144: Is this supposed to be a 200?

DonnieBLT (changes requested):
See comment @ATHARVA-GAWAS

website/tests.py
ATHARVA-GAWAS: Let me check again

ATHARVA-GAWAS: Since the test if for whether the form is valid or not

DonnieBLT: The test should be if the rate limit was hit

Issue #1808, opened 2024-02-21
Bug: Missing Pages


ATHARVA-GAWAS:
Example

![image](https://github.com/OWASP/BLT/assets/111885892/f1bc0cdb-6034-47f8-9d8f-6b9dd067523b)

ATHARVA-GAWAS:
We may close this issue now as it is merged

PR #1809, opened 2024-02-21
Coming Soon Pages Added

ATHARVA-GAWAS:
Issue Solved
#1808 

image


website/templates/bacon.html
```diff
@@ -0,0 +1,21 @@
+{% extends "base.html" %}
```
arkid15r, line 1: These pages looks identical. If that's the case I recommend following the DRY principle and crate a "coming soon" template page to use it as a base for the new pages.

ATHARVA-GAWAS: Ok, let me check

ATHARVA-GAWAS:
Done, made a file comingsoon.html and extended it to other pages

arkid15r (commented):
I think the naming convention suggests "coming_soon.html".

Other than that looks good 👍 (I didn't review the page's from UI perspective)


upd: correct the extension.

AtmegaBuzz:
ArcoLinux_2024-02-22_00-05-17

For UI i would suggest making it something like this

ATHARVA-GAWAS:
Better?

image

DonnieBLT (changes requested):
Can you please make it look exactly like the suggestion?

ATHARVA-GAWAS:
Ok, i will do it

ATHARVA-GAWAS:
Is this fine?

image

DonnieBLT (changes requested):
One tweak

ATHARVA-GAWAS:
Ok

ATHARVA-GAWAS:
Actually base.html is essential, since without it, the entire structure of the page gets messed up

ATHARVA-GAWAS:
Changes done

DonnieBLT (approved):

Issue #1210, opened 2023-04-07
Project: Invites & Referrals


solo-daemon:
UPDATE (2024/1/27): Have it be a template that will be sent by the user - lets do that to avoid any potential misuse - we can have a link in the email that will give the user a reward when the friend signs up.

- [ ] #581

solo-daemon:
This can be achieved after #1209

letsintegreat:
I don't think we should send the invites via our server. Even with API rate limiting, one can use a set of proxies to target email spam through us. Instead, we can draft an invite message with link and send an intent to user's email and messaging clients. Another plus is that - this way, the receiver will receive the invite via a trusted source, then the receiver is more likely to accept the invite.

AtmegaBuzz:
I think it's ok to use email for sending invite from server. We were thinking of integrating invite friend with gmail, The main idea is that we create a token and send it in email to register and if registered with that email then the sender will automatically will be rewarded with BACON coins and this Invite api will be used in mobile app also. 
And schedulers aren't required for sending emails, we can use multi threading to achieve background tasks and thread will closed after email send or fail.

DonnieBLT:
I like the idea of having it be a template that will be sent by the user - lets do that to avoid any potential misuse - we can have a link in the email that will give the user a reward when the friend signs up.

kr-2003:
> I like the idea of having it be a template that will be sent by the user - lets do that to avoid any potential misuse - we can have a link in the email that will give the user a reward when the friend signs up.

@DonnieBLT In what form, does the user gets rewarded??

Issue #581, opened 2018-03-01
Task: Integrate a referral program


fredfalcon:
- [ ] Task: Build the landing page
- [ ] Task: Integrate the referral link into the email
- [ ] Task: Track clicks to the referral email
- [ ] Task: Track user signups from the referral link

https://github.com/byteweaver/django-referral

![image](https://user-images.githubusercontent.com/7475382/87330516-af50b880-c506-11ea-9d6e-f7fef99cf4ac.png)

PR #1813, opened 2024-02-22
Project: Invites & Referrals (implemented)

JisanAR03:
After evaluating our needs for implementing a referral system in our application, decided not to use the django-referral library. The primary reasons are its lack of support for Django 2.0 and several errors that hindered integration.

**What We Did:**

- Model Update: We have updated the InviteFriend model to include additional fields such as sender, recipients, referral_code. These enhancements are aimed at tracking invitations more efficiently and enabling the application to monitor the sign-up status of invited users.

- Security Measures: To safeguard our system against potential spam and abuse, we've implemented several security-related validations. These measures ensure that our invitation system remains secure and is used as intended, promoting a healthy and spam-free environment.

- Invitation Page Update: Minor improvements were made to the invitation page, enhancing its usability and visual appeal.

- Reward System Adjustments: We've refined our reward system. Previously, users could earn rewards for merely sending out invitations. Now, rewards are granted only when the invited users complete the signup process successfully. This change aims to encourage genuine invitations and active participation within our community.

**Conclusion**:
These updates bring our referral system in line with modern standards, ensuring it's both secure and user-friendly. By making these changes, we're not only improving the user experience but also reinforcing the integrity of our platform.

This PR fix issue #1210 #581 .
@DonnieBLT  sir, could you please review this PR and share your feedback? Thank you very much.😊


DonnieBLT (changes requested):
This looks good can you please check the thread we decided we want to do something where we generate the referral code and track the referral. However we won't send the email out to use the user will send the email out with their own email service, so what we could do is show the email and then have a butt to "send email "that will automatically fill out the subject in the body from the URL. Using an email href

JisanAR03:
> This looks good can you please check the thread we decided we want to do something where we generate the referral code and track the referral. However we won't send the email out to use the user will send the email out with their own email service, so what we could do is show the email and then have a butt to "send email "that will automatically fill out the subject in the body from the URL. Using an email href

Ok, sir, sure. and will push the update after completing this one

JisanAR03:
delete this one for conflict issue

PR #1823, opened 2024-02-24
Update pre-commit: enable builtin literals check

arkid15r:
Enables `check-builtin-literals `pre-commit` check.


fredfalcon (approved):

DonnieBLT (approved):

DonnieBLT:
image
looks like we have it enabled but would not be able to delete a branch from a different repository

Issue #1820, opened 2024-02-24
on the user's profile, make the tip button open the QR code for the BCH address and remove the PayPal form.


DonnieBLT:

Sarthak5598:
![Generate Bitcoin QR Code - Opera 24-02-2024 15_31_35](https://github.com/OWASP/BLT/assets/108613416/6c2282f3-0022-4d8b-ac2f-1d8ed49e50a9)
![Generate Bitcoin QR Code - Opera 24-02-2024 15_31_51](https://github.com/OWASP/BLT/assets/108613416/5085781b-8767-4db7-94f2-85ffb82ca68a)
![Generate Bitcoin QR Code - Opera 24-02-2024 15_31_41](https://github.com/OWASP/BLT/assets/108613416/38e46bfc-a38c-4c87-b6eb-9eafb75ae7e9)
@DonnieBLT 
this is the improvement so far 
I want your recommendations on how to get receiver's address , Should I add a column on user's profile and then we can directly fetch addresses eg->receiver's address = user.BCH_adrs

DonnieBLT:
It can just be a modal on the page. No need for the form as the user has already entered the address

Sarthak5598:
Okay sir

Sarthak5598:
![BLT - Opera 24-02-2024 21_44_40](https://github.com/OWASP/BLT/assets/108613416/050d7375-5ac7-4deb-91fc-bebb51bafd0b)
![BLT - Opera 24-02-2024 21_44_50](https://github.com/OWASP/BLT/assets/108613416/827abbba-0594-4e70-9758-6f5ae2a03072)
Now, the send tip button directly shows a QR
Do you want me to make any changes?

DonnieBLT:
![image](https://github.com/OWASP/BLT/assets/128622481/5f66570a-3e97-41d0-ba1a-66545459925c) yes can you please have the QR code appear as something like this and underneath please show the BCH address and the letters BCH

Sarthak5598:
![BLT - Opera 24-02-2024 21_44_40](https://github.com/OWASP/BLT/assets/108613416/10fd1227-4f12-467b-ae15-cff217de2ee4)
![BLT - Opera 26-02-2024 19_07_03](https://github.com/OWASP/BLT/assets/108613416/d6aa38cc-7745-4b9a-895e-447e5b9b3e0b)
what do you think?
@DonnieBLT

DonnieBLT:
[...]

PR #1846, opened 2024-02-28
on the user's profile, make the tip button open the QR code for the BCH address and remove the PayPal form. #1820

Sarthak5598:
#1820

### DIFF
--- a/blt/urls.py
+++ b/blt/urls.py
@@ -416,7 +416,6 @@ urlpatterns = [
     path("company/", include("company.urls")),
     path("sponsor/", website.views.sponsor_view, name="sponsor"),
     path("companies/", DomainListView.as_view(), name="domain_lists"),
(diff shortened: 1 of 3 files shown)


Sarthak5598:
@DonnieBLT Please check this out , I have removed the django function and used only js . So, the issue related to import qr code wont happen again !! 
thanks for the help @AtmegaBuzz

AtmegaBuzz:
Awesome, I tested it and its working as expected. just increase the height and width of img tag to 400 
like this QR Code
rest looks good

AtmegaBuzz:
BTW @DonnieBLT is there any option for users to add bch address from profile ?

Sarthak5598:
I made the change @AtmegaBuzz

DonnieBLT (approved):

DonnieBLT:
> BTW @DonnieBLT is there any option for users to add bch address from profile ?

There might not be.

Issue #1819, opened 2024-02-24
add a small tip button on the leaderboard that opens the QR code for the users BCH address


DonnieBLT:

Sarthak5598:
@DonnieBLT Do you want a button here for QR?
![image_2024-02-26_195040331](https://github.com/OWASP/BLT/assets/108613416/1f423711-2dff-4e4e-8170-fd1eb6506e7e)

DonnieBLT:
Yes "send a tip" and pop up the qr.

Sarthak5598:
![image_2024-02-27_075904440](https://github.com/OWASP/BLT/assets/108613416/42d6d3fa-8337-4e7c-b880-de1df6ecc0c0)
what do you think ?
@DonnieBLT

PR #1849, opened 2024-02-28
add a small tip button on the leaderboard that opens the QR code for the users BCH address #1819

Sarthak5598:
#1819

### DIFF
--- a/website/templates/_leaderboard_widget.html
+++ b/website/templates/_leaderboard_widget.html
@@ -48,7 +48,7 @@
                 </span>
             </a>
             <div>
-                <button type="button" id="openModal" class="btn btn-danger w-full" style="color: black;">Send a Tip</button>
+                <button type="button" id="openModal" class="btn btn-danger w-full">Send a Tip</button>
                 <dialog id="dialog">
                     <img id="qr-image" alt="QR Code" height="400" width="400">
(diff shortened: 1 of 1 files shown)


HanilJain (commented):
@Sarthak5598  **To add Tip button** 
suggestion : to add tip button you can replace the rank span and replace it with the tip button.
and rank can be displayed over the avatar of the user and really small so that it does not cover the avatar but enough readable

DonnieBLT:
That sounds like a good idea. Can we see a screenshot of this when it's ready please

Sarthak5598:
Something like this thanks to @HanilJain for the idea 
Screenshot 2024-02-29 010252

Sarthak5598:
I still have 1 issue the tip button is also taking us to user profile as its under anchor tag and if I move it out of anchor tag it goes to next line

HanilJain:
> Something like this thanks to @HanilJain for the idea Screenshot 2024-02-29 010252

@Sarthak5598 Awesome. it would look even better if you add colour to rank like gold , silver and bronze.

HanilJain:
> I still have 1 issue the tip button is also taking us to user profile as its under anchor tag and if I move it out of anchor tag it goes to next line

make that a div and tip in a anchor tag

Sarthak5598:
it worked @HanilJain thanks a lot !!!

DonnieBLT (approved):

PR #1851, opened 2024-02-28
Github and Google login test

ATHARVA-GAWAS:


website/tests.py
```diff
@@ -145,3 +151,53 @@ def test_on_hide(self):
...
+                raise ValueError("Unexpected URL name")
+
+        def mock_build_absolute_uri(uri):
+            return "http://example.com" + uri
+
+        with patch("django.urls.reverse", side_effect=mock_reverse):
+            with patch("django.http.HttpRequest.build_absolute_uri", side_effect=mock_build_absolute_uri):
+                callback_url = view.callback_url
+
+        self.assertEqual(callback_url, "http://example.com/accounts/google/login/callback/")
```
DonnieBLT, line 178: This doesn't seem to be testing anything? What is example.com why are we testing their site?  Also what issue is this solving?

DonnieBLT (changes requested):
See comment

website/tests.py
ATHARVA-GAWAS: How can i directly use BLT domain??

ATHARVA-GAWAS: Need to figure that out

PR #1857, opened 2024-02-29
Issue 1819 Issue with button visiblity in light mode

Sarthak5598:


Sarthak5598:
@DonnieBLT please merge!

DonnieBLT:
Can you send a screenshot of what it looks like with these changes thanks

Sarthak5598:
Screenshot 2024-02-29 222109

DonnieBLT:
Thanks it seems like the QR code has no way to close it. Can you please add a close button to the model, and if the address is none or empty, let's not show the QR code.

Sarthak5598:
Okay ! But if user doesn't have the address the tip button would look like its not working ,Maybe we can show a message instead of showing QR ?
and also If possible can you make it a separate task ? 
Because I have to make changes in both the tip button

DonnieBLT (approved):

PR #1957, opened 2024-03-06
Adding CVE number when reporting a bug  

HanilJain:
### Added #1927 

While reporting bug now CVE number can be attached with the bug report. 
With an additional feature of validating check.

Screenshot 2024-03-07 040440

Screenshot 2024-03-07 040801

PR #1974, opened 2024-03-08
Suggested changes in CVE ID 

HanilJain:
Suggested changes from #1957
by @arkid15r


HanilJain:
@arkid15r Thank you for your suggestions, 
I have made all the changes you suggested. Are there more changes required

DonnieBLT (approved):
Thanks for the update

PR #1981, opened 2024-03-09
Step 1: Create the Django model Define a Django model named 'Monitor' with fields such as 'url', 'keyword', 'last_checked_time', 'status', and 'user'. #1897

Sarthak5598:
#1897


DonnieBLT (changes requested):
Can you also add the created and modified fields as well as the migration and the class for the admin?

DonnieBLT (changes requested):
Please also add the form to create an entry as well as the created / modified field names.

Sarthak5598:
sorry, I forgot about this, I will make the changes !!

Sarthak5598:
image

DonnieBLT (changes requested):
Almost there we also need a created and modified field. Please see the other table for an example.

website/views.py
```diff
@@ -2438,6 +2445,19 @@ def create_wallet(request):
...
+    if request.method == "POST":
+        form = MonitorForm(request.POST)
+        if form.is_valid():
+            monitor = form.save(commit=False)
+            monitor.user = request.user  # Assuming you have a logged-in user
+            monitor.save()
+            # Redirect to a success page or render a success message
+    else:
+        form = MonitorForm()
+    return render(request, "Moniter.html", {"form": form})
```
arkid15r, line 2458:
```suggestion
    return render(request, "monitor.html", {"form": form})
```

arkid15r, line 2458: Do you use this view anywhere?

website/views.py
```diff
@@ -4257,3 +4277,15 @@ def get_context_data(self, **kwargs):
...
+def create_monitor(request):
+    if request.method == "POST":
+        form = MonitorForm(request.POST)
+        if form.is_valid():
+            form.save()  # This will create a new Monitor instance
+            messages.success(request, "Form submitted successfully!")
+    else:
+        form = MonitorForm()
+
+    return render(request, "Monitor.html", {"form": form})
```
arkid15r, line 4291: Based on the  current `templates` directory naming convention:
```suggestion
    return render(request, "monitor.html", {"form": form})
```

website/views.py
Sarthak5598: Ohh! forgot about that thanks

DonnieBLT (approved):

PR #1996, opened 2024-03-11
suggested tip amount based on cve number if a CVE is attached to an issue report

HanilJain:
# Feature Added
###  Issue #1928 

- **Added CVE Score** : If a CVE ID is attached to issue report it automatically fetches the CVE Score from this API Service https://services.nvd.nist.gov/rest/json/cves/2.0?cveId and saves to models 

- **functioning of Send tip feature** :  send tip feature was not working on /issue2/ page in this PR it was made functional 

- **Suggestion of tip** : On the basis of CVE Score, tip and severity is decided 
    -  cve_score >= 9.0  : critical : $50
    -  cve_score >= 7.0  : high : $20
    - cve_score >= 4.0 : medium : $10
    - cve_score >= 0.0 : low : $5 

 
Screenshot 2024-03-11 153934


website/templates/issue2.html
```diff
@@ -536,6 +610,70 @@ <h2 class="mb-2 text-xl font-bold leading-tight text-gray-900 ">{{ issue.descrip
...
+        let suggestedTip;
+        if (severityLevel === "critical"){
+            suggestedTip = "$50";
+        } else if (severityLevel === "high") {
+            suggestedTip = "$20";
+        } else if (severityLevel === "medium") {
+            suggestedTip = "$10";
+        } else {
+            suggestedTip = "$5";
+        }
```
arkid15r, line 634: It seems you can combine `severityLevel` and `suggestedTip` setting logic:
```suggestion
        let severityLevel;
        let suggestedTip;

        if (cve_score >= 9.0) {
            severityLevel = "critical";
            suggestedTip = "$50";
        } else if (cve_score >= 7.0) {
            severityLevel = "high";
            suggestedTip = "$20";
        } else if (cve_score >= 4.0) {
            severityLevel = "medium";
            suggestedTip = "$10";
        } else {
            severityLevel = "low";
            suggestedTip = "$5";
        }
```

HanilJain: @arkid15r ohh I missed that , ThanksYou

website/models.py
```diff
@@ -312,6 +313,22 @@ def remove_user(self):
     def get_absolute_url(self):
         return "/issue/" + str(self.id)
 
+    def get_cve_score(self):
+        if self.cve_id:
```
arkid15r, line 317: Consider checking `if self.cve_id is None` first:
```suggestion
        if self.cve_id is None:
            return None
```
to save an indent below

website/models.py
```diff
@@ -312,6 +313,22 @@ def remove_user(self):
...
+                response = requests.get(url).json()
+                results = response["resultsPerPage"]
+                if results != 0:
+                    metrics = response["vulnerabilities"][0]["cve"]["metrics"]
+                    if metrics:
+                        cvss_metric_v = next(iter(metrics))
+                        score = metrics[cvss_metric_v][0]["cvssData"]["baseScore"]
+                        self.score = score
+                    return score
+            except Exception as e:
```
arkid15r, line 329: Could you narrow down the expected exceptions set? It's not a good practice to catch broad exception.

website/models.py
```diff
@@ -312,6 +313,22 @@ def remove_user(self):
...
+            try:
+                url = "https://services.nvd.nist.gov/rest/json/cves/2.0?cveId=%s" % (self.cve_id)
+                response = requests.get(url).json()
+                results = response["resultsPerPage"]
+                if results != 0:
+                    metrics = response["vulnerabilities"][0]["cve"]["metrics"]
+                    if metrics:
+                        cvss_metric_v = next(iter(metrics))
+                        score = metrics[cvss_metric_v][0]["cvssData"]["baseScore"]
+                        self.score = score
```
arkid15r, line 327: It should either return the `score` or set it, not both. Based on the function name I'd go with return option.

HanilJain:
@arkid15r  I have made all the suggested changes

website/models.py
```diff
@@ -312,6 +313,26 @@ def remove_user(self):
     def get_absolute_url(self):
         return "/issue/" + str(self.id)
 
+    def save_cve_score(self):
```
arkid15r, line 316: I liked the old name better. The current name doesn't reflect what the method does (technically it doesn't save the score but rather gets it via API and puts into `self.score`).

website/models.py
```diff
@@ -312,6 +313,26 @@ def remove_user(self):
...
+                if metrics:
+                    cvss_metric_v = next(iter(metrics))
+                    score = metrics[cvss_metric_v][0]["cvssData"]["baseScore"]
+                    self.score = score
+        except requests.exceptions.HTTPError as e:
+            print(e)
+            return None
+        except requests.exceptions.ReadTimeout as e:
+            print(e)
+            return None
```
arkid15r, line 334: Could you group these exceptions:
```suggestion
        except (requests.exceptions.HTTPError, requests.exceptions.ReadTimeout) as e:
            print(e)
            return None
```

website/models.py
HanilJain: ohkay !!

HanilJain:
@arkid15r thanks for the suggestion, please check

website/migrations/0087_issue_cve_score.py
arkid15r, line 1: I've just realized we already have `score`. Do you use both fields to save the same CVE base score value?

HanilJain: @arkid15r  damnn it was typo it meant to be 
`self.cve_score = score `
thanks for mentioning it I have made the changes

website/models.py
```diff
@@ -312,6 +313,23 @@ def remove_user(self):
...
+        try:
+            url = "https://services.nvd.nist.gov/rest/json/cves/2.0?cveId=%s" % (self.cve_id)
+            response = requests.get(url).json()
+            results = response["resultsPerPage"]
+            if results != 0:
+                metrics = response["vulnerabilities"][0]["cve"]["metrics"]
+                if metrics:
+                    cvss_metric_v = next(iter(metrics))
+                    score = metrics[cvss_metric_v][0]["cvssData"]["baseScore"]
+                    self.cve_score = score
```
arkid15r, line 328: My original concern was regarding redundant setting `self.cve_score` here as you also doing it in the views.py L884. Consider just returning it as a result of `Issue::get_cve_score()`. Other option would be rename it to `set_cve_score()`.

arkid15r (commented):
Yes, looks better!

DonnieBLT (changes requested):
one small tweak from me

website/models.py
HanilJain: Done 👍

website/templates/issue2.html
```diff
@@ -536,6 +610,65 @@ <h2 class="mb-2 text-xl font-bold leading-tight text-gray-900 ">{{ issue.descrip
...
+        { name: 'Bitcoin', address: btc },
+        { name: 'Ethereum', address: eth },
+        { name: 'Bitcoin Cash', address: bch }
+        ];
+        let currentIndex = 0;
+
+        const updateSlide = () => {
+            let { name, address } = cryptoOptions[currentIndex];
+            if (address == '' || address == 'None') {
+                address = "qr5yccf7j4dpjekyz3vpawgaarl352n7yv5d5mtzzc";
```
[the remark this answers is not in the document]
HanilJain: image

website/models.py
```diff
@@ -312,6 +313,23 @@ def remove_user(self):
...
+        try:
+            url = "https://services.nvd.nist.gov/rest/json/cves/2.0?cveId=%s" % (self.cve_id)
+            response = requests.get(url).json()
+            results = response["resultsPerPage"]
+            if results != 0:
+                metrics = response["vulnerabilities"][0]["cve"]["metrics"]
+                if metrics:
+                    cvss_metric_v = next(iter(metrics))
+                    score = metrics[cvss_metric_v][0]["cvssData"]["baseScore"]
+                    return score
```
arkid15r, line 328: Optional: another optimization you could use is 
```suggestion
                    return metrics[cvss_metric_v][0]["cvssData"]["baseScore"]
```
right away as you're using `score` variable just once.

arkid15r (commented):
Awesome, thank you @HanilJain!

Looks good!

website/models.py
HanilJain: ohh yess thanks

DonnieBLT (approved):

PR #1999, opened 2024-03-13
solve the image showing condittion

### LINKED ISSUE #1998: homepage have showing more then one image in each bug and for some image the title isn't showing
screenshot: 
image

JisanAR03:
before : 
image

after : 
image

@DonnieBLT sir , please review this one .
fix #1998


DonnieBLT (changes requested):
some of the older bugs have bug.image (double check this) so lets show that if it exists else show screenshot.  Also if there is more than one image, we'd like to show subtle left and right arrows so they can be scrolled through as a carousel.

JisanAR03:
> some of the older bugs have bug.image (double check this) so lets show that if it exists else show screenshot. Also if there is more than one image, we'd like to show subtle left and right arrows so they can be scrolled through as a carousel.

ok sure sir

DonnieBLT (approved):

PR #2085, opened 2024-05-09
Issue #2043

Sarthak5598:


website/views.py
```diff
@@ -4399,3 +4403,87 @@ def deletions(request):
...
+
+    byte_io = io.BytesIO()
+    image.save(byte_io, format="PNG")
+    byte_io.seek(0)
+
+    return HttpResponse(byte_io, content_type="image/png")
+
+
+def SaveBiddingData(request):
+    form = BidForm()
```
arkid15r, line 4423: It seems you create the form but never use it for input [validation](https://docs.djangoproject.com/en/5.0/ref/forms/validation/).

website/views.py
```diff
@@ -4399,3 +4403,87 @@ def deletions(request):
...
+    if request.method == "POST":
+        data = json.loads(request.body)
+        issue_url = data.get("issue_url")
+        bid = Bid.objects.filter(issue_url=issue_url).order_by("-created_at").first()
+        if bid is not None:
+            bid_amount = bid.amount
+            time_left = bid.created_at - datetime.now(timezone.utc)
+            user = bid.user
+            date = bid.created_at
+            status = bid.status
```
arkid15r, line 4452: These values can be inlined directly into the `JsonResponse`

website/views.py
```diff
@@ -4399,3 +4403,87 @@ def deletions(request):
...
+                }
+            )
+        else:
+            return JsonResponse({"error": "Bid not found"}, status=404)
+    else:
+        return JsonResponse({"error": "Method not allowed"}, status=405)
+
+
+def submit_pr(request):
+    form = BidForm()
```
arkid15r, line 4469: Same here -- your form is unused and you create `Bid` instance directly from the POST request data.

arkid15r (commented):
Could you address validation part when you get a chance?

DonnieBLT (changes requested):
a couple more changes requested

DonnieBLT (approved):

PR #2370, opened 2024-06-25
Asynchronous Notification 

HanilJain:
#### Issue Addressed 

#2213 
#2214
#2215
#2216
#2217
#2218
#2221

#### How it works 

- Notification is created in model of notification_app
- new_home frontend fetch api Notification.
- websockets in header.html file handles the notification coming and managed by redis-server
- redis-server handles the cache and channel serving the notification to webscoket at port 6379
- when user click on notification it gets deleted from database 

#### Image 

image

#### Note 
There can be issue with combability with pythonanywhere so before merigng please confirm it once.
This PR just for code correction and suggestion.


arkid15r (changes requested):
@HanilJain could you remove the unrelated 5000+ files (virtual environment?) from the PR and make the commit history clean by squashing the 150+ commits?
Thank you!

Screenshot 2024-06-25 at 9 20 13 AM

arkid15r (commented):
Please also make sure the pre-commit runs green locally before pushing it to GitHub.

PR #2382, opened 2024-06-29
fix company account creating issue

### LINKED ISSUE #2380: Fix the company part error
when we go to "http://127.0.0.1:8000/company/"

now showing in frontend : 
image

we are expecting : 
image

and after solving the path issue now the another error is when submit the company creating form : 
image

JisanAR03:
for make it work we have to run "python manage.py migrate"

fixed #2380


website/models.py
```diff
@@ -47,6 +47,7 @@ def generate_uuid_for_company(apps, schema_editor):
 
 
 class Company(models.Model):
+    company_id = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)
```
arkid15r, line 50: Maybe we should create an [abstract base model](https://docs.djangoproject.com/en/5.0/topics/db/models/#abstract-base-classes) with `uuid` fields and use it for inheritance when we need to have uuid field (the name would be the same for all models)?

JisanAR03: @arkid15r sir, I update the code

website/models.py
```diff
@@ -46,7 +46,14 @@ def generate_uuid_for_company(apps, schema_editor):
         obj.save()
 
 
-class Company(models.Model):
+class UUIDModel(models.Model):
+    company_id = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)
```
arkid15r, line 50: Almost there -- as I mentioned you would need to have a unified name for the field. This way we would be able to use it in any model. If you name it `company_id` it fits companies context only

```suggestion
    uuid = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)
```

arkid15r: It's also a good idea to extract `UUIDModel` to a separate module accessible from any of our Django apps.

JisanAR03: @arkid15r sir ,  we need company_id field in company model for solve the error , now if we name it "uuid" the error not gonna fix . 
(basically the logic is coded by previous developer ,for now we have to consider this field not as a common field cause backend part is already done by using company_id so now maybe we don't need 
     `class UUIDModel(models.Model):
    company_id = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)

    class Meta:
        abstract = True`
 
 
 so for the company,  we can just add :
    `company_id = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)`
    
    
    now waiting for confirmation from you for make it like before

arkid15r: > @arkid15r sir , we need company_id field in company model for solve the error , now if we name it "uuid" the error not gonna fix . (basically the logic is coded by previous developer ,for now we have to consider this field not as a common field cause backend part is already done by using company_id so now maybe we don't need

Let's take a look at the logic you've just mentioned. Is there an easy way to change it to lookup the entity based on `uuid` fiedl instead of `company_id`?

JisanAR03: > Let's take a look at the logic you've just mentioned. Is there an easy way to change it to lookup the entity based on `uuid` fiedl instead of `company_id`?

@arkid15r  sir ,those logic are huge , I have already checked and if I start changing on those stuff that will take huge time

arkid15r: Apparently, I'm not gonna make you refactor a bunch of code just to have a unified field name. However, I'd like to understand what the logic you're referring to is. As I couldn't figure out any details neither from the PR nor from the [related issue](https://github.com/OWASP-BLT/BLT/issues/2380). Is it about `CompanyViewSet`?

JisanAR03: @arkid15r ok sir now I update the issue and explain everything actually what I'm gonna fix

JisanAR03: @arkid15r , can you please check the description from : #2380

arkid15r: Well, something is wrong there for sure. The `company` should not have a field `company_id`  in the first place (it's already in the company context so simple `id` would be enough -- and actually `id` field is already there).

It seems that we need to fix the `Company` form submission logic instead of introducing the new field.
Unfortunately I don't have time to troubleshoot the #2380 today. I'm going to look into this later.

JisanAR03: @arkid15r , ok then sir I will try to fix the error my own , then will ask suggestion from you

DonnieBLT (changes requested):
I don't understand why we have a UU ID

JisanAR03:
> I don't understand why we have a UU ID

@DonnieBLT , basically the previous code store uuid in company_id .that's why I also use uuid . if uuid is not needed then I will fix this error by using primary key

JisanAR03:
@DonnieBLT , @arkid15r  . maybe we should have use company_id with a uuid . cause as basis of screenshot. there is already a function which make a uuid for each data , and our production application have few company so how can we remove the uuid without losing the company data?

image

AtmegaBuzz:
company_id was included during prev gsoc and somebody removed it causing the company dashboard to fail, i've used company_id in url so @JisanAR03 you have to fix the part where we are making the model change and separating the company id into another model. Company id should be unique and will shown in url while accessing companies and UUID is best we can use @DonnieBLT.

PR #2389, opened 2024-07-01
commented out auto labellling functionality

Sarthak5598:


DonnieBLT (dismissed):

arkid15r (commented):
Would reverting the commit instead of commenting everything out work better here?

In general it's not a good practice to have commented code in the codebase.

Sarthak5598:
@arkid15r Sure If you feel thats the best , the old pr can be reverted

DonnieBLT (approved):

PR #2392, opened 2024-07-02
fix the cache error which also occurred error on run-test

JisanAR03:
the change will fix the cache error from home page and also fix the run-test error too

### DIFF
--- a/website/views.py
+++ b/website/views.py
@@ -24,6 +24,7 @@ import tweepy
 # from django_cron import CronJobBase, Schedule
 from allauth.account.models import EmailAddress
 from allauth.account.signals import user_logged_in, user_signed_up
+from django.contrib.auth.signals import user_logged_out
 from allauth.socialaccount.providers.facebook.views import FacebookOAuth2Adapter
 from allauth.socialaccount.providers.github.views import GitHubOAuth2Adapter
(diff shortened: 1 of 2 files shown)


JisanAR03:
@DonnieBLT , @arkid15r ,, there is a cache related issue I have notice , for that reason the home page not work perfectly . and also showing "run test" error in every PR . now It's fixed . can you please review for merge this update

DonnieBLT (approved):
Awesome thank you!

website/views.py
```diff
@@ -272,6 +272,11 @@ def _cache_middleware(request, *args, **kwargs):
     return _decorator
 
 
+@receiver(user_logged_out)
+def handle_user_logged_out(request, user, **kwargs):
+    cache.clear()
```
arkid15r, line 277: Why do you clear the entire cache upon this event?

JisanAR03: yes sir, should i just remove the cache_key?

arkid15r: I'm not sure what issue you fixed with that but it seem an overkill to me in any case as the **entire** cache is being invalidated upon any user signin/signup/logout event, right?

JisanAR03: yeah right sir , I will fix this up for sure

PR #2396, opened 2024-07-03
update the cacheing

JisanAR03:
@arkid15r  sir, can you please check this commit . update the caching


arkid15r (commented):
Maybe instead we shouldn't cache anonymous user responses at all?

arkid15r:
There is a standard Django way to [cache a page](https://docs.djangoproject.com/en/5.0/topics/cache/#the-per-view-cache). In `newhome` there is some extra logic for authenticated users email check/verification that can be a middleware and cover all routes instead of just the home page.

JisanAR03:
@arkid15r , any suggestion for me ?

arkid15r:
The point of not having authenticated user related logic is that we could use `cache_page` freely -- the content would be the same for everyone. 

Also it may be a good idea to start from the other end in terms of page loading optimization -- index [Issue::is_hidden](https://github.com/OWASP-BLT/BLT/blob/f18932312f7341eb425b229a90c0a13e54dc87ea/website/views.py#L298) or  [Point::created](https://github.com/OWASP-BLT/BLT/blob/f18932312f7341eb425b229a90c0a13e54dc87ea/website/views.py#L311) or make other changes to speed up the view logic.

arkid15r:
> @arkid15r , any suggestion for me ?

No, not at this time. I believe it's better to leave it for review to @DonnieBLT

JisanAR03:
@arkid15r ok sure. and also with that I can optimize the home page to . so I'm thinking to create a issue and start optimize the home page . 


@DonnieBLT sir , should i do that ? and can you please review this PR too.

JisanAR03:
@DonnieBLT , now this PR is just update the login cache if user trigger "signin, signout , or logout " any of this ,, my last pr was clear all the cache now this PR only clear the login related cache

DonnieBLT (approved):

JisanAR03:
> Awesome thanks! Feel free to create another issue to address the other concerns.



> Awesome thanks! Feel free to create another issue to address the other concerns.



> Awesome thanks! Feel free to create another issue to address the other concerns.

@DonnieBLT  Thank you so much

PR #2397, opened 2024-07-03
fix 2380 issue

JisanAR03:
**Impact:**
- Existing company details in the main application may be lost if we proceed with the new model changes.
- Modifying the endpoint to incorporate the new company ID will require significant time and effort, as all related endpoints will need to be updated.

**Proposed Solution:**
Rather than making extensive changes to the model and endpoints, we can resolve the issue by simply adding the company ID to the current company model. This solution is straightforward and will preserve the existing company details and functionality.

**Conclusion:**
The company ID is already a built feature used in various views. Therefore, we should avoid changing the model structure and opt for the simpler solution of adding the company ID to the existing model.

@DonnieBLT , @AtmegaBuzz , @arkid15r .. can you please review and any suggestion about what i did wrong ?


arkid15r (commented):
I guess you know my opinion based on previous PR review.  If you can't use existing `company.id` add another field. I don't think it should be named `company_id` though as it's already `Company` model.

In general I don't mind have this as a temporary solution. It seems the codebase already has many of them.

JisanAR03:
> I guess you know my opinion based on previous PR review. If you can't use existing `company.id` add another field. I don't think it should be named `company_id` though as it's already `Company` model.
> 
> In general I don't mind have this as a temporary solution. It seems the codebase already has many of them.

@arkid15r , actually sir , i did that cause initially I can't start working on company stuff without fixing this error  , that's why I make this changes

JisanAR03:
> Not sure if my comments are being seen. Please do not use UUID. Let's use a company_slug and the built in integer id as unique. id can be the primary key (it should already be built in) and slug is unique. There is one record we can delete so don't worry about the data.

ok cool , if we don't have to worry about the data , then I can complete whole stuff my own with help of @arkid15r mentorship . thanks a lot @DonnieBLT

PR #2401, opened 2024-07-04
update company endpoint with companyID

JisanAR03:
remove uuid from company model and also update all the endpoint for match with new model

@DonnieBLT , now company_slug using the built in integer id as unique key


arkid15r (commented):
Looks good, thanks for refactoring this!

I suggest renaming  `id/company` params in views definitions to `company_id` where applicable for better readability (e.g. [here](https://github.com/OWASP-BLT/BLT/blob/841d088457539a0b106311d224cd907ba0803d40/company/views.py#L805))

DonnieBLT (approved):
LGTM

PR #2411, opened 2024-07-07
HomePage: Scroll Based Pagination added

### LINKED ISSUE #2311: Implement JavaScript code for scroll-based pagination.

CodeWithBishal:
This closes #2311 and #2312 #2313 

### Desktop Behaviour

As the user scrolls down and reaches the end of the first page, JavaScript sends a request to fetch and display the contents of the next page. This process continues until all pages have been loaded.

### Mobile Behaviour

On mobile devices, the leaderboard appears just below the list of bugs. Implementing scroll-based pagination here would require users to scroll through all pages, potentially wasting server resources. Therefore, for mobile devices, I have kept the manual pagination using page numbers.

### DIFF
--- a/website/templates/new_home.html
+++ b/website/templates/new_home.html
@@ -2,22 +2,60 @@
 {% block content %}
     {% include "includes/sidenav.html" %}
     <style>
-        .hero-bugs-container {
-            grid-template-columns: repeat(auto-fill, minmax(225px, 1fr));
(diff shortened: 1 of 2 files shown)


DonnieBLT (approved):
This looks good, we decided to not use auto load so we will keep this closed and reopen if we need it,

PR #2495, opened 2024-07-23
add ability to block ip addresses that are submitting spam #1938

Sarthak5598:
#1938


DonnieBLT (changes requested):
a few comments

DonnieBLT (changes requested):

DonnieBLT (changes requested):

blt/middleware/ip_restrict.py
```diff
@@ -0,0 +1,35 @@
...
+    def blocked_ips(self):
+        blocked_ips = cache.get("blocked_ips")
+        if blocked_ips is None:
+            # Retrieve all blocked IP addresses from the database
+            blocked_addresses = BlockedIP.objects.values_list("address", flat=True)
+            blocked_ips = set(blocked_addresses)
+            cache.set("blocked_ips", blocked_ips, timeout=86400)
+        return blocked_ips
+
+    def __call__(self, request):
```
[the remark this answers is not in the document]
Sarthak5598: Isnt blocking the user agent itself is ineffective as it will block it for every user with that user agent and the spammer can always use other device , why would we need it ?

blt/middleware/count_ip_requests.py
```diff
@@ -0,0 +1,29 @@
...
+            ip_record = IP.objects.filter(address=ip).first()
+
+            if ip_record:
+                # Update the count and save
+                ip_record.user_agent_string = parsed_agent
+                ip_record.count += 1
+                ip_record.save()
+                print(f"Updated IP record with address {ip}. New count: {ip_record.count}")
+            else:
+                print(f"IP {ip} does not exist in the database.")
```
arkid15r, line 29: Please note this is a print, not logging. And it's going to be executed for **every** request (the logging could be quite excessive).

blt/middleware/ip_restrict.py
```diff
@@ -0,0 +1,35 @@
...
+        ip = request.META.get("REMOTE_ADDR")
+
+        if ip:
+            blocked_ips = self.blocked_ips()
+            if ip in blocked_ips:
+                return HttpResponseForbidden(
+                    "Your IP address is restricted from accessing this site."
+                )
```
arkid15r, line 32:
```suggestion
        if ip and ip in self.blocked_ips():
            return HttpResponseForbidden("Your IP address is restricted from accessing this site.")
```

blt/middleware/ip_restrict.py
```diff
@@ -0,0 +1,35 @@
...
+
+        response = self.get_response(request)
+        return response
```
arkid15r, line 35:
```suggestion
        return self.get_response(request)
```

website/models.py
```diff
@@ -658,6 +661,16 @@ def __str__(self):
         return f"Q: {self.question} | A: {self.answer} at {self.timestamp}"
 
 
+class BlockedIP(models.Model):
```
arkid15r, line 664: It looks like you could use a flag of `IP` model, something like `is_blocked` and toggle it as needed instead of creating duplicates for every blocked IP address.

PR #2631, opened 2024-08-12
Enhancement: Profile page

CodeWithBishal:
Fix some overflow issues in the design of profile page
added an edit button to edit user's crypto address

### DIFF
--- a/website/static/js/validate-btc.js
+++ b/website/static/js/validate-btc.js
@@ -28,7 +28,7 @@ function validateBCH(address) {
      * Params: BCH Address
      * ***/
     if (address == null || address == "" || address == " ") {
-        return false;
+        return "empty";
     }
     if (address.startsWith("bitcoincash:")) {
         address = address.slice(12);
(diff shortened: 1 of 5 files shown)


CodeWithBishal:
@DonnieBLT can we keep both of the options? because this PR also contains design adjustments for the profile page to fix overflow issue

CodeWithBishal:
Yep it should work with the current profile edit option

CodeWithBishal:
@DonnieBLT after the merge I just need to make a few changes and "edit profile" should work fine with this PR

CodeWithBishal:
right now, in my PR the `update_bch_address` is expecting a JSON payload which is not the case in the "edit profile" page, need to adjust this and it should work just fine

CodeWithBishal:
ok

CodeWithBishal:
I have made the required changes, please review @DonnieBLT

DonnieBLT (approved):

PR #2656, opened 2024-08-18
feature to Recommend User on profile page

npxpatel:
- User can recommend others on their profile
- Created a page to make recommendations

fixes issue -#2632

https://github.com/user-attachments/assets/dc35a90f-1c87-467e-a65b-45e4f5b0cde2


npxpatel:
@DonnieBLT any feedback or suggestions you might have .

npxpatel:
So instead of a new route "/recommend" , do we have to put up a FORM in the profile page to recommend a user ? If so, where do we display the list of recommendations by a user.

npxpatel:
Is this method of showing recommendations made by a user correct ( in the video of this PR) ,  So the changes to make is putting that recommendation FORM in the profile page ?

tsu-ki:
@npxpatel can i work on this issue, if you're okay with that?

PR #2673, opened 2024-08-21
BLT Private Project

HanilJain:
### description
BLT private project have Private Issue Page where all the Private Issue are listed and viewership can be managed.
request made by any user and people who have access are listed which can be granted or declined the access.
User could request access for Private Issue with message and the owner will notified by the mail , If access is granted the user will be notified through email and can view the Private Issue.

### Feature

- Request Access for Private Issue
- Email Notification on Request Request and approval
- viewership management of Private Issue
- dedicated page for Private Issue management
- Well Tested and security management

### workflow

image

### Screenshots

image
image
image
image


DonnieBLT (commented):
Thanks! I think we want to go about this a different way but we can count this as done. Thank you.

Issue #2229, opened 2024-06-01
Allow users to report suspicious bug reports.


DonnieBLT:

snigdhachoudhury:
@coderabbitai is this issue being worked upon currently?

Issue #2227, opened 2024-06-01
Enable reviewers and domain managers to verify reports.


DonnieBLT:

Krishiv-Mahajan:
@coderabbitai explain the issue please

snigdhachoudhury:
@coderabbitai  answer these questions
1. Should organization admins (not just domain managers) be able to verify issues for their organization's domains?
2. Should there be a minimum reputation/points threshold before a user can become a verifier?
3. Can the issue reporter verify their own issue, or should that be blocked?
4. Should the verification badge show on issue list pages (not just detail page)?
5. Should domain managers see a "Verify" button in the domain dashboard alongside each issue?
6. Should there be a bulk-verify action on issue list pages for managers?
7. Should unverifying an issue require a reason/comment?
8. Should there be a verification history/audit log showing who verified/unverified and when?
9. Can a verified issue be edited, or should editing reset verification status?
10. Should domain managers be notified when issues in their domain need verification?
11. Should the verification notification include the verifier's username?
12. Should the API endpoint be at [/api/v1/issue/verify/<id>/] or [/organization/verify_issue/<id>/]?
13. Should the API return full issue data or just [{verified: true/false}]?

Issue #2226, opened 2024-06-01
Implement spam check logic within Django views.


DonnieBLT:

igennova:
@coderabbitai how can i solve this?

Issue #2223, opened 2024-06-01
Create a dataset for spam detection.


DonnieBLT:

igennova:
@coderabbit what can be the solution here?

DonnieBLT:
We can now use an LLM for spam detection what are some low cost options?

PR #2854, opened 2024-11-03
Project Spam Detection 

HanilJain:
### Issues Solved
#2223 
#2224
#2225 
#2226
#2227 
#2228 
#2229 

https://github.com/user-attachments/assets/0536976c-61ae-4daa-b3ee-6401178d4eca

We have trained an ML model using a publicly available spam dataset, which has been integrated within an API. This model predicts whether the description or title of an issue is spam or not. If flagged as spam, the issue is marked accordingly in the system and then proceeds to a human review layer for further validation.


DonnieBLT (changes requested):
Can you please add tests to make sure this is working?

HanilJain:
@DonnieBLT @AtmegaBuzz @arkid15r Could you please review this PR.

DonnieBLT (changes requested):

PR #2931, opened 2024-11-15
Activity feed page creation

### LINKED ISSUE #2915: Add new page for activity feed
Add activity for new issue added and others

krrish-sehgal:
Fixes: #2915

### Model

The `Activity` model consists of the following fields:
- `user`: The user who performed the action (ForeignKey to User).
- `action_type`: The type of action performed (e.g., created, updated, deleted).
- `timestamp`: The time when the activity occurred.
- `title`: A short title for the activity (e.g., issue title or blog title).
- `description`: A detailed description of the activity.
- `image`: An optional image associated with the activity.
- `related_object`: A reference to the object involved in the activity (ForeignKey to models like Issue, Blog, BugBounty, or IPReport).

### Signals

- **Pre-save signal**: Triggered when an object is about to be saved, to log any updates or changes that are about to occur.
- **Post-save signal**: Fired after an object is saved to the database, creating a new `Activity` entry to capture the event.
- **Post-delete signal**: Activated when an object is deleted, logging the deletion event in the activity feed.

### How It Handles Key Features

The system handles the following key actions related to issues, blogs, bug bounties, and IP reports:

1. **Capturing Creation, updation , deletion of Issues, Blogs, Bug Bounties, and IP Reports**:

2. **Displaying Activities**:
   - In the frontend, activities are displayed in a **feed** format, with each activity card showing the user who performed the action, the type of action (create, update, or delete), and the timestamp.

https://github.com/user-attachments/assets/91b8528c-5fc0-4a5f-8555-784bdd9071a2


krrish-sehgal:
@DonnieBLT  , Please Feel free to give feedback on the functionality....

DonnieBLT (changes requested):
Can you please see my feedback? Also, maybe it might be better to use Django-activity-feed what do you think?

krrish-sehgal:
Thanks for your feedback,I have taken into account all your points, I'll start making the changes right away.

> Can you please see my feedback? Also, maybe it might be better to use Django-activity-feed what do you think?
It has some limitations, particularly around performance at scale and domain-specific flexibility. Since its generic to its actor-verb-object model,it might not fully accommodate the aspects of different features of our projects like IP reports, bug hunts, issues, blogs and anything we further add, without significant customization. Additionally, since it's third-party library, debugging can be challenging—I've encountered unexpected signals while working with it before, and resolving such issues is often more complex compared to custom logic.

DonnieBLT (dismissed):

DonnieBLT (dismissed):

DonnieBLT (dismissed):

krrish-sehgal:
@DonnieBLT , its good to merge now👍

DonnieBLT (approved):

PR #2939, opened 2024-11-16
fix: fixed hyperlink issue in signup email

### LINKED ISSUE #2934: Email for verification always ends up in spam and contains long verification link.
Tried creating an account with two different accounts, ends up in spam
Image

  
A suggestion would be to add an href tag instead of copying that long string containing the token  
and pasting it in the browser. 
This would be better from a UX standpoint imo.

1010varun:
fixes: #2934


1010varun:
So what should I can do to make it a link

1010varun:
Yess, yesterday I tried the same thing but I didn't get the function from where this is being called. Can you please help me in this?

1010varun:
Done with adding the .html template. Got the solution in allauth docs

DonnieBLT (dismissed):

1010varun:
@DonnieBLT just forgot to run pre-commit previously, can you please approve once one

DonnieBLT (approved):

Issue #2936, opened 2024-11-16
Figma and GitHub links floating in the corner on the company register page.


apoorvapendse:
![Image](https://github.com/user-attachments/assets/5b679cb1-875a-4af7-93eb-fd73c155c1ac)

tsu-ki:
/unassign
created a PR for this issue, it only needs to be reviewed now

PR #2942, opened 2024-11-17
Corrected Figma and GitHub links floating in the corner on the company register page(Issue #2936)

tsu-ki:
Removed the absolute styling and added a flex-based layout structure where the flexbox ensures that the FIgma and github links stay at the bottom even when the content is shorter than the viewport
Screenshot 2024-11-17 at 3 07 31 PM


tsu-ki:
#2936 added flex-based structure

DonnieBLT (approved):

PR #2943, opened 2024-11-17
fix: fix the position of logout modal on page

### LINKED ISSUE #2938: Logout button confirmation must be at the center of the screen and should appear as modal instead of being at the top left corner.
Image

1010varun:
fixes: #2938 and #2965

I added the flex to make the modal center of the page.


1010varun:
But @DonnieBLT, don't you think doing this will affect the current alignment of other pages. I think these are correctly aligned currently.

DonnieBLT (changes requested):
Please have this change be at the base level, so it applies to all the pages in allauth

1010varun:
Surely le me do that

DonnieBLT (changes requested):
This should work with only one file the base file

1010varun:
@DonnieBLT there is a change only in base.html, in logout.html I have made a whole new modal for logout that's why there are changes in that file.

previous logout:
image


changes in logout:
1. made a box and shadow
2. changed button color to match the theme
3. Make text to be in center of modal

Screenshot from 2024-11-20 00-13-49

DonnieBLT (changes requested):
please make it the base.html only - the base.html should go in the allauth or account folder so it overrides the allauth one

harsh3dev:
@1010varun still working on this one?

1010varun:
Yess

1010varun:
@DonnieBLT, I think there is some misunderstanding in this PR there are changes in two file:
1. base.html 
2. Logout.html

the changes in base.html make all the auth pages in the enter of screen and changes in logout.html are done so to create a new modal for logout. Which is another issue(#2965). If you want I can
 make changes in base,HTML and then raise another PR for logout modal changes.

I am attaching some ss for new auth screens
Screenshot from 2024-11-20 00-11-04
Screenshot from 2024-11-20 00-13-49
Screenshot from 2024-11-20 00-10-56

1010varun:
@DonnieBLT can you kindly look into this PR

DonnieBLT (approved):

PR #2959, opened 2024-11-21
Profile View Count Logic and UI for #2633

SahilDhillon21:
Related to issue #2633

Summary of changes:

- **Added views field** to the UserProfile model to track profile views.
- **Updated the UserProfileDetailView** to increment the view count every time the profile is accessed.
- **Displayed view count** in the profile template.

profile-views-ui

**Possible improvements**
The view count gets updated on every reload. The same thing happens with issue view count as well. We can add a session or IP-based approach to prevent incrementing the view count on every refresh.


DonnieBLT (approved):

DonnieBLT (changes requested):
please merge the migration and also please use the IP table to get the count for the profile page - you can check the recent code created for the readme badge we have that shows the stats.

DonnieBLT (dismissed):

DonnieBLT (approved):

PR #2966, opened 2024-11-23
Fix org name restriction, reference organization while starting timer.

### LINKED ISSUE #2851: allow the sizzle time tracker to be attached to an org and have settings on the org to display time tracking publicly or not.

apoorvapendse:
This is a PR in reference to issue 2851.
The bug inducing check of needing to match org name and domain name is removed with this PR.
Fixes #2851
Also the user can now optionally reference an org they want to allocate the current sizzle timelog towards with the help of the organization URL.

Right now, only the user who created the timer can see which organization they allocated their time logs.
image


DonnieBLT (changes requested):
It looks good just two changes requested

DonnieBLT (dismissed):

DonnieBLT (approved):

PR #3027, opened 2024-11-30
Improved search dropdown icon usage

tsu-ki:
Fixed Issue: https://github.com/OWASP-BLT/BLT/issues/2984

Before(Iphone XR dimensions): 
Screenshot 2024-12-01 at 1 20 10 AM

After(Iphone XR dimensions): 
Screenshot 2024-12-01 at 1 20 36 AM

### DIFF
--- a/website/templates/includes/header.html
+++ b/website/templates/includes/header.html
@@ -141,7 +141,7 @@
                                placeholder="Search"
                                class="text-xl lg:text-base lg:ps-9 lg:scale-[1.4] w-[70px] lg:w-auto h-16 placeholder-[#9CA3AF] bg-[#EDEDED] outline-none p-5 font-semibold rounded-l-2xl">
                         <select name="category"
-                                class="text-xl lg:text-base lg:scale-[1.4] w-[70px] lg:w-auto h-16 placeholder-[#9CA3AF] bg-[#ededed] outline-none p-5 font-semibold rounded-r-2xl text-[#9CA3AF]">
+                                class="text-xl lg:text-base lg:scale-[1.4] flex-grow h-16 placeholder-[#9CA3AF] bg-[#ededed] outline-none p-5 font-semibold rounded-r-2xl text-[#9CA3AF]">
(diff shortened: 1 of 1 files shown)


krrish-sehgal (commented):
How about we add icons for organisations and search when the width is reduced on a mobile screen, instead of displaying the whole text?

tsu-ki:
> How about we add icons for organisations and search when the width is reduced on a mobile screen, instead of displaying the whole text?

@krrish-sehgal We'll have to implement it for the other option values as well ( projects, users, tags etc), otherwise just using an icon for organization might look odd.
what do you say?

krrish-sehgal:
Arent the other options confined in a sidenav? so displaying text for them is not an issue, but the header looks messy with full words "search" and "organisations" ,we can just have icons for them instead of full words (and you can take the same organisations icon being used in sidenav)

tsu-ki:
> Since the other options are already confined within the sidenav, displaying text for them shouldn’t pose an issue. However, the header feels cluttered with full words like 'Search' and 'Organisations.' Perhaps we could simplify it by using just icons instead, similar to the 'Organisations' icon already being used in the
> 
> you can ofcourse wait for @DonnieBLT sir, to ensure this aligns with the overall design direction. Thanks.

I agree with you on this one, I too believe the header sections looks very cluttered and using icons would simplify things; but on another note, for a new user it's easier navigate using the dropdown names instead of icons

krrish-sehgal:
By default only icons will be displayed and as soon as the organisation icon is clicked , drop down with full length names is displayed.

tsu-ki:
> How about ,by default only icons will be displayed and as soon as the organisation icon is clicked , drop down with full length names is displayed? and whenever a new thing is selected we can show their icon.
> 
> You can use the icons from font awesome library for orgs, users, projects which are pre used in sidenav and then you can use <i class="fas fa-language" <i class="fas fa-tags" for language and tags.

`<option>` element cannot directly render icons, so I'll have to use an EventListener and add some css to attain this. I'll make the changes and show the results here

PR #3057, opened 2024-12-04
Update the activity feed to be anonymous

### LINKED ISSUE #3053: When an issue is marked as anonymous, also update the activity feed to be anonymous

SahilDhillon21:
Fixes #3053

Whenever the anonymous button is clicked, after removing the user the corresponding activity object is also modified to default to an anonymous user.
Additionally, I have included an extra "Report Anonymously" option while creating the issue itself in case user directly wants to report it anonymously, removing an extra step.


DonnieBLT (changes requested):
This looks good, can you please also create a test for it? Thank you!

DonnieBLT (approved):

PR #3098, opened 2024-12-12
Emoji and yesterday's work

### LINKED ISSUE #3097: For the sizzle check-in show yesterday’s goal and have a checkbox if the goal was accomplished, also have another field for how you’re feeling today where the user can choose any emoji

SahilDhillon21:
Fixes #3097

checkin-reports
detailed_checkin_report
sizzle_form


DonnieBLT (changes requested):
Can you please make the emoji dynamic and maybe use a library?

DonnieBLT (approved):

PR #3141, opened 2024-12-20
Mandatory issue screenshot

### LINKED ISSUE #2228: Create review queues with permissions for reviewers.

SahilDhillon21:
Partially fixes #2228

I have shifted the validation before the object is created, in case something like this doesn't happen:

1. Submit a request without screenshots
2. The object gets created
3. The validation runs and tries to delete the object
4. If there's any error during deletion or a race condition, the object might persist

However, some bots have uploaded random white images like this: 
image

So shall we implement a review queue system?


DonnieBLT (dismissed):
Looks good, please check the comment for one potential fix.

DonnieBLT (approved):

PR #3238, opened 2025-01-17
Slack field in project.

### LINKED ISSUE #3198: Add a new field slack_channel to projects
Also create a csv of all OWASP slack channels with an import script.

krrish-sehgal:
fixes #3198
Sir i have fixed the issue. 
let me know if any more changes are needed.


krrish-sehgal:
@DonnieBLT sir , can you once review this and let me know if any more additions to be done for this ,or should resolve the run tests.

krrish-sehgal:
Sure sir .

DonnieBLT (changes requested):
Let’s change the approach to have a new model SlackChannels, then pull all of the slack channels that match #project-*  It’s been a while since the other pool request has activity so you can copy any code from there and use it on this PR as I mentioned you will both get credit

PR #3257, opened 2025-01-23
Trademark fetching and results display

### LINKED ISSUE #1925: On the company detail page have a section called Trademarks - and it will list all trademarks for that company

SahilDhillon21:
Fixes #1925

I have made use of the same RapidAPI URL present in the `check_trademarks.py` command.

Details page:

https://github.com/user-attachments/assets/ec17f74d-2709-44c0-b764-8b9429ce741a


DonnieBLT (changes requested):
it would be good to store these in a model so we don't call the api each time

DonnieBLT (dismissed):

DonnieBLT (approved):

PR #3268, opened 2025-01-24
added sidebar to search page and added the ability to search and see everything

### LINKED ISSUE #3261: add sidebar to search page and add the ability to search and see everything

tsu-ki:
Closes: https://github.com/OWASP-BLT/BLT/issues/3261

https://github.com/user-attachments/assets/23e4ab2e-2e0b-4232-a6cf-bcef0185cc2c


tsu-ki:
@DonnieBLT sir this PR is ready to be reviewed, please do suggest if there are any viable changes to be made. thank you!

DonnieBLT (changes requested):
Can you keep the organization link we want the ability to search all plus the organization. Also, can you show what the data will look like on the return screen?

tsu-ki:
> Can you keep the organization link we want the ability to search all plus the organization. Also, can you show what the data will look like on the return screen?

https://github.com/user-attachments/assets/7b8e91e7-9f18-4d05-99d9-7154875db907

I've set `All` data-value as default so that user doesn't have to change between different values for finding something, they ofcourse can be specific but it's not a necessity anymore. I've attached a test run video, please review this, thank you!

DonnieBLT (approved):

PR #3286, opened 2025-01-27
Script to link exisiting orgs with domains and make domain view safe

### LINKED ISSUE #3279: Domain page is not showing any information
Image

SahilDhillon21:
Fixes #3279

1. Added a management command to **link Domain instances to Organization entities**. 
2. **Replaced get_or_404** with error handling that allows page rendering even if trademarks fail to load.
3. Modified the get_name property on Domain model to ensure the URL has a scheme and safely extract the domain name


tsu-ki:
@SahilDhillon21 does this script handle cases where domains are different from organisation's name?
while creating gsoc organisation json file, I noticed that many orgs have domains dissimilar to their names. If this isn't taken into consideration, it might cause errors

tsu-ki:
> > @SahilDhillon21 does this script handle cases where domains are different from organisation's name? while creating gsoc organisation json file, I noticed that many orgs have domains dissimilar to their names. If this isn't taken into consideration, it might cause errors
> 
> Hopefully their domain name at least contains the org name, if not then we'll have to add more methods to link them Can you please send me the json so that I can test this script and possibly improve it to handle different cases

sure thing, if you're able to implement linking  gsoc org's domain to their blt profile page, please do let me know as I'm having trouble with that right now. I've shared the file with you on slack

DonnieBLT (approved):

PR #3386, opened 2025-02-11
Automated command for deletion of unverified users

### LINKED ISSUE #3387: Django Management command for deleting unverified users

tsu-ki:
Closes: https://github.com/OWASP-BLT/BLT/issues/3387 

**Test run without making changes**

`python manage.py cleanup_unverified_users --dry-run`

**Actually run the cleanup**

`python manage.py cleanup_unverified_users`

image

### DIFF
--- a/blt/settings.py
+++ b/blt/settings.py
@@ -570,4 +570,15 @@ CHANNEL_LAYERS = {
     },
 }
 
+USER_CLEANUP_SETTINGS = {
+    "REMINDER_HOURS": 24,
+    "DELETE_HOURS": 48,
+    "PROTECTED_EMAIL_DOMAINS": [
+        "owasp.org",
+        DOMAIN_NAME,
+    ],
+    "BACKUP_ENABLED": True,
+    "BACKUP_PATH": os.path.join(MEDIA_ROOT, "user_backups"),
+}
+
 ORD_SERVER_URL = os.getenv("ORD_SERVER_URL", "http://localhost:9001")  # Default to local for development
--- a/website/management/commands/cleanup_unverified_users.py
+++ b/website/management/commands/cleanup_unverified_users.py
@@ -1,0 +1,240 @@
+import json
+import logging
+import os
+from datetime import timedelta
+
+from allauth.account.models import EmailAddress
+from django.conf import settings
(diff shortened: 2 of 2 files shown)


CodeWithBishal:
can we have atleast a week and 2 reminder emails before it automatically deleting an account?
`python manage.py cleanup_unverified_users`
will it remove all the unverified users or just a particular user?

if all the users will be removed, will it check the accounts that are getting deleted got the mails & it's past 1 week since the first reminder.

- First reminder to be sent 1 week before deletion
- Second reminder 24 hours before deletion

tsu-ki:
> can we have atleast a week and 2 reminder emails before it automatically deleting an account? `python manage.py cleanup_unverified_users` will it remove all the unverified users or just a particular user?
> 
> if all the users will be removed, will it check the accounts that are getting deleted got the mails & it's past 1 week since the first reminder and remove only those who fulfils the criteria?
> 
> * First reminder to be sent 1 week before deletion
> * Second reminder 24 hours before deletion

currently implementation sends a reminder mail 24 hours after the account has been created. If the user doesn't verifies their account between 24 hours to 48 hours window, their account gets deleted after that. 
`python manage.py cleanup_unverified_users` removes **all** unverified users except admins, staff, superusers, members from an organization or if the user's email is from `owasp.org` domain. 

I haven't add the cross verification function yet. I'll add this aswell, thanks for your suggestion

DonnieBLT (changes requested):
I think for the first time we run this we need a different approach.

tsu-ki:
> I think for the first time we run this we need a different approach.

@DonnieBLT sir, that approach should I follow for the first execution?

tsu-ki:
> > > I think for the first time we run this we need a different approach.
> > 
> > 
> > @DonnieBLT sir, that approach should I follow for the first execution?
> 
> Print a list . The following 100 users have not verified their email. Then have a flag auto_delete=false

okay, so in the first execution there won't be any automatic deletion but rather just print a list of unverified users?

tsu-ki:
> I just realized we don’t require email verification so we’d have to turn that on first then wait a month or so. Let’s revisit this later

sure thing, I'll keep this PR until then. thank you for updating me!

PR #3437, opened 2025-02-17
Added allauth base templete (#3415)

### LINKED ISSUE #3415: fix all the allauth pages
there is a way to add one base file in allauth / accounts and have it provide a base for all files like https://blt.owasp.org/accounts/social/connections/

Kaushik-Shahare:
Closes #3415 

image


DonnieBLT (changes requested):
This should only be one file changed, adding the base template and then overriding the styles

DonnieBLT (changes requested):
you should be able to extend the base.html and then add the extra_head block to inject the style tag and then override the form, input etc elements

PR #3679, opened 2025-02-21
feat: enhance URL validation for OWASP compliance checks to prevent S…

### LINKED ISSUE #3680: Fix code scanning alert
Tracking issue for:
- [ ] https://github.com/OWASP-BLT/BLT/security/code-scanning/190

DonnieBLT:
…SRF and improve security

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved URL validation to enhance security and provide clearer error messages for URL submissions.
- **Refactor**
  - Made internal code improvements to boost maintainability while keeping user-facing interactions unchanged.


gojo-satorou-v7 (changes requested):
Instead of validating the url here too for SSRF we can use rebuild_safe_url from utils.py and then implement the logic later.
Also according to me this should be a high priority issue, I had also flagged this problem earlier.

@CodeRabbit check if my above solution is efficient/implementable or not also check if this PR should be in high priority in #3942  or not.

PR #3685, opened 2025-02-21
Style hook and migration

### LINKED ISSUE #3416: remove bootstrap and make sure the whole site is using tailwind and there is no css

SahilDhillon21:
Fixes #3416

A pre-commit hook has been added that checks for <style> in HTML files and blocks the commit if found. It can be bypassed using an arg --allow-styles for special cases.

Thanks a lot for the CSS file view, I'll go over them and modify the files to use tailwind without changing the final UI


SahilDhillon21:
The status is stuck for some reason, and it's showing conflicts even after copying the changes. Please look at https://github.com/OWASP-BLT/BLT/pull/3906/ for now

PR #3925, opened 2025-03-11
Fixed the page loading takes too much time and implemented lazy loading instead of that on statistic charts on repo details page

### LINKED ISSUE #3897: add statistic charts to the repo detail page
Image

swaparup36:
https://github.com/user-attachments/assets/efd5c6e7-8c9a-42e8-8ac6-a9744972a49f

It resolves #3897 

I just implemented lazy loading of the statistic chart. That reduced the page's loading time. As the calculation and data gathering for the charts is time-consuming, it takes some time to load after the page is rendered as shown in the video. 

## Summary by CodeRabbit

- **New Features**
  - Introduced a dynamic endpoint to provide live repository activity data.
- **Refactor**
  - Updated the repository statistics section to always display charts with "Loading…" placeholders that update dynamically, ensuring a more responsive and intuitive user experience.


CodeWithBishal:
@swaparup36 
- What actions could be taken to reduce the page loading time
- re-commit with pre-commit

swaparup36:
@CodeWithBishal 

As we need to get very detailed data of 2 months for making the chart, so we have to call the API of GitHub again and again to get all the required data. So, a possible solution could be caching these data into redis, it might help the page to load faster

PR #4064, opened 2025-03-27
Added PopUp for Email Confirmation

### LINKED ISSUE #4049: User signup email verification
when signed up. The username is taken and if the email was incorrect, there is no way to edit and change it.

https://blt.owasp.org/issue/5508 found by Anonymous at url: http://blt.owasp.org

Himanshu-639:
fixes #4049

This PR implement a Popup for email confirmation before sending the verification email to user.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the sign-up experience by adding an email confirmation popup. Users receive a prompt to confirm or edit their email before finalizing registration.
  
- **Tests**
	- Improved the automation tests for the sign-up process to ensure more reliable element interactions during user registration.

### DIFF
--- a/website/templates/account/signup.html
+++ b/website/templates/account/signup.html
@@ -174,6 +174,26 @@
                 </form>
             </div>
         </div>
+        <!-- Popup HTML Structure -->
+        <div id="emailConfirmationPopup"
+             class="hidden fixed inset-0 bg-gray-800 bg-opacity-75 flex items-center justify-center">
+            <div class="bg-white rounded-lg p-6 shadow-lg">
+                <p>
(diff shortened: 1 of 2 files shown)


gojo-satorou-v7 (changes requested):
@Himanshu-639 The issue was that when an user signs up and had mistakenly entered the wrong email address the username gets taken away and there should be a way to edit the mail after the verification has been sent. However your changes just ask for confirmation, an user can obviously make a mistake here too.

Please fix this and remember to handle the edge cases, like after the email change the old confirmation link should not work.

Issue #3856, opened 2025-03-08
Create a private page on the organization that shows all slack apps installed and their commands


DonnieBLT:

igennova:
Do you mean that only an admin can access the private page?

igennova:
@coderabbit code the perfect solution for this

Ravitheja1289-dot:
@coderabbitai explain the problem and tell me if someone has already given out a pr for this

PR #4467, opened 2025-07-30
Create private page on the organization shows all slack apps and their commands

Nitish-Naik:
feature: #3856 
Created a private page on the organization that shows all slack apps installed and their commands.
Example: /blt help


DonnieBLT:
seems this was picked up in a new PR

PR #5124, opened 2025-11-28
Added unique_email constraint and a script to remove duplicates

### LINKED ISSUE #1978: Make the email field unique and create a script that removes users with duplicate emails (keep the first)
This should run as part of the migration so there is no error when the migration runs (there will be no existing duplicates)

sidd190:
Fixes #1978 

Screenshot from 2025-11-28 20-11-20
Screenshot from 2025-11-28 20-11-52

# Changes Made

## 1. Migration: `website/migrations/0259_make_email_unique.py`

* Identifies users with **duplicate non-empty email addresses**
* Keeps the **first user (lowest ID)** for each duplicate email
* **Deletes all other users** with the same email
* Adds a **partial unique index** on the `email` field
* Allows **multiple empty emails**
* Enforces **uniqueness for non-empty emails**
* Uses **proper logging** instead of print statements
* Wrapped in an **atomic transaction** to ensure rollback on failure

---

## 2. Management Command: `website/management/commands/analyze_duplicate_emails.py`

A pre-migration analysis tool to preview which users will be deleted:

* Shows **total count** of duplicates and users to be removed
* `--detailed` flag displays **specific users kept vs deleted**
* Helps administrators review impact **before running the migration**

---

# Technical Details

## Migration Flow

1. Query for duplicate emails
2. Delete duplicates (keep first user by ID)
3. Create unique index:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS auth_user_email_unique
ON auth_user (email)
WHERE email != '';
```

### Key Features

* **Partial unique index** allows multiple users with empty emails (common for social auth)
* PostgreSQL-specific syntax: `WHERE email != ''`
* Idempotent SQL via **IF NOT EXISTS**
* Proper logging for **audit trail**
* Clear warnings about **irreversibility**

---

# Usage

### Preview what will be deleted

```bash
python manage.py analyze_duplicate_emails --detailed
```

### Run migration

```bash
python manage.py migrate website 0259
```

---

# Why This Approach?

* Prevents future **duplicate email registrations**
* Maintains **data integrity**
* Keeps the **oldest user account** as the canonical record
* Handles edge cases involving **empty emails** (e.g., social auth users)

---

# Breaking Changes

⚠️ **Warning:**
This migration **permanently deletes users** with duplicate emails.
Related objects will be **cascade-deleted** according to their foreign key constraints.
Always run the **analysis command first** in production.

## Summary by CodeRabbit

* **New Features**
  * Added a CLI command to scan and report duplicate non-empty email addresses, with an optional detailed mode showing which accounts would be kept or removed and summary statistics.

* **Chores**
  * Enforced uniqueness for non-empty user emails via a data migration: duplicate accounts will be removed while preserving the earliest account per email and a conditional unique index will be created; deletions are irreversible.


Pritz395 (changes requested):
Hey! Found some critical issues with the migration:

1. Database Compatibility Issue
The migration uses PostgreSQL-specific SQL syntax (WHERE email != '' at line 78-80) that will fail on SQLite/MySQL. Since we're using SQLite in dev (according to settings.py), this migration won't work.

2. Missing NULL Email Handling
The code only excludes empty strings (email=""), but Django's User.email can also be None. This means None values will incorrectly show up in the duplicate check (lines 22 and 33).

3. No Transaction Wrapper
The user deletions happen in a loop without transaction protection (lines 38-44). If one deletion fails partway through, we'll end up with inconsistent data.

4. Performance Issue
Deleting users one-by-one in a loop is really slow. With many duplicates, this could take a while since we're making N individual delete() calls instead of using bulk delete.

Not blockers but worth noting: there's no check if kept_user is None (unlikely edge case), and the management command has the same NULL email issue.

sidd190:
> Hey! Found some critical issues with the migration:
> 
> 1. Database Compatibility Issue
>    The migration uses PostgreSQL-specific SQL syntax (WHERE email != '' at line 78-80) that will fail on SQLite/MySQL. Since we're using SQLite in dev (according to settings.py), this migration won't work.
> 2. Missing NULL Email Handling
>    The code only excludes empty strings (email=""), but Django's User.email can also be None. This means None values will incorrectly show up in the duplicate check (lines 22 and 33).
> 3. No Transaction Wrapper
>    The user deletions happen in a loop without transaction protection (lines 38-44). If one deletion fails partway through, we'll end up with inconsistent data.
> 4. Performance Issue
>    Deleting users one-by-one in a loop is really slow. With many duplicates, this could take a while since we're making N individual delete() calls instead of using bulk delete.
> 
> Not blockers but worth noting: there's no check if kept_user is None (unlikely edge case), and the management command has the same NULL email issue.

Ahhh, I'll fix the top 3, though we're using PostgreSQL in dev. (default is sqlite and for local testing), but I've made it database agnostic. Also for the last one, performance, we can still handle a little more time taken, but we want Django signals to fire since in data migrations, correctness > speed and bulk delete would induce potential risks.

PR #5127, opened 2025-11-28
feat : slack channel integration in projects

### LINKED ISSUE #3198: Add a new field slack_channel to projects

Nachiket-Roy:
# Slack Channel Integration & Project Linking

Closes: #3198 

Screenshot

### 1. Create & Update SlackChannel Objects
- CSV and API imports now correctly create or update SlackChannel entries using:
  - `slack_id` → `channel_id`
  - `slack_channel` → `name` 
  - `slack_url` → `slack_url`
- Preserves existing project links when updating channels

### 2. Link SlackChannel → Project (ForeignKey Integration)
- New matching system links Slack channels to projects using:
  - Exact case-insensitive name match
  - Partial match fallback (safe only when a single candidate)
- Linked channels update both:
  - `SlackChannel.project`
  - Legacy fields: `project.slack`, `project.slack_channel`, `project.slack_id`

### 3. Backward Compatibility Preserved
- Even though SlackChannel is now the primary store, legacy Project fields remain updated:
  - `project.slack`
  - `project.slack_channel` 
  - `project.slack_id`
- This ensures:
  - Older templates still work
  - API responses expecting these fields remain compatible
  - No breaking changes for existing consumers

### 4. Improved Matching Logic
- New helper methods implement:
  - Normalization of Slack channel → project names
  - Exact match detection
  - Safe no-match behavior with warnings

### 5. Detailed Logging Coverage
- Clear logs for:
  - Exact matches
  - Partial matches
  - Unmatched channels
  - Missing required CSV fields
- Makes debugging CSV problems significantly easier

### 6. UI – Slack Channel Badge in Project Cards
- Adds first SlackChannel badge to project list view
- Links to the channel's URL
- Gracefully hidden when no channel exists

### 7. CSV-Based Import Support
- CSV is now fully supported alongside Slack API import
- CSV import follows the same normalization & linking logic as the API
- Ideal for bulk admin imports and testing

### 8. Tests Added
Comprehensive tests for:
- SlackChannel creation
- SlackChannel update
- Project linking (exact + partial)
- Normalization logic
- Backward compatibility updates
- API fetch success and failure cases
- Missing fields handling
- Unmatched channels logging
- Empty CSV behavior

## Summary by CodeRabbit

* **New Features**
  * Slack channel info shown on project list and detail pages (name, member count, privacy badge, link).
  * Import tool accepts CSV or API, normalizes channel data, matches/links channels to projects, updates legacy project Slack fields, improved logging and no-op behavior.

* **Tests**
  * Added tests for import creation/updating, exact/partial matching, legacy-field updates, and unmatched-channel handling.

### DIFF
--- a/website/management/commands/import_slack_channel.py
+++ b/website/management/commands/import_slack_channel.py
@@ -48,19 +48,24 @@ class Command(BaseCommand):
         Try to find a Project instance that corresponds to a given Slack channel name.
         Returns (project, match_type) where match_type is 'exact', 'partial', or None.
         """
-        normalized_name = self.normalize_project_name(channel_name)
 
-        # 1) Exact (case-insensitive) match on name
-        exact_qs = Project.objects.filter(name__iexact=normalized_name)
+        # 1) Try exact match with original channel name (case-insensitive)
+        exact_qs = Project.objects.filter(name__iexact=channel_name)
         if exact_qs.exists():
             return exact_qs.first(), "exact"
 
-        # 2) Partial match using icontains on normalized_name (lowercase)
+        # 2) Try exact match with normalized name
+        normalized_name = self.normalize_project_name(channel_name)
+        exact_normalized_qs = Project.objects.filter(name__iexact=normalized_name)
+        if exact_normalized_qs.exists():
+            return exact_normalized_qs.first(), "exact"
+
+        # 3) Partial match using icontains on normalized_name
         partial_qs = Project.objects.filter(name__icontains=normalized_name)
         if partial_qs.count() == 1:
             return partial_qs.first(), "partial"
 
-        # 3) No reliable match
+        # 4) No reliable match
         return None, None
 
     def fetch_from_api(self):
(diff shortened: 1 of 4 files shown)


Pritz395 (changes requested):
Hey, overall looks good but a few changes could be made:-

- IntegrityError risk (Line 180) — We're setting project.slack_id = channel_id without checking if another project already has that ID. Since slack_id has unique=True, this'll crash with an IntegrityError if there's a duplicate.

**Does Not Block Approval but could be addressed as well since it's a straightforward fix:** 
- Performance problem (Line 63) — Project.objects.all() loads every project into memory. This'll get slow and memory-heavy as we scale. We should filter at the queryset level instead.
- Race condition (Lines 70-71) — We're calling count() to check for one result, then first() to grab it. Between these calls, the data could change. Better to use a single query.

DonnieBLT:
This was already added.

PR #5133, opened 2025-11-30
feat: Add GitHub integration fields to Issue model

satvikpathak:
This PR adds comprehensive test coverage for the GitHub integration fields on the Issue model and improves data integrity by using `PositiveIntegerField` for comment counts.

## GitHub Integration Fields Added

This PR works with the following 4 GitHub integration fields that were added to the Issue model:

* **`github_comment_count`** (`PositiveIntegerField`) - Number of comments on the GitHub issue
* **`github_state`** (`CharField`) - Current state of the GitHub issue (open/closed) as reported by GitHub API
* **`github_fetch_status`** (`BooleanField`) - Whether GitHub data was successfully fetched from the API
* **`github_last_fetched_at`** (`DateTimeField`) - Timestamp of when GitHub data was last fetched (tracks freshness of GitHub data)

### Model Improvements
* Changed `github_comment_count` from `IntegerField` to `PositiveIntegerField` to prevent negative values and ensure data integrity.
* Updated migration (0260) to use `PositiveIntegerField` to match model definition.

### Test Coverage Added
* **`test_issue_github_last_fetched_at`** - Verifies timestamp tracking for GitHub data fetches:
    * Default value is `None` for new issues.
    * Timestamp is set when GitHub data is successfully fetched.
    * Timestamp is updated on subsequent fetches.
* **`test_issue_github_state_invalid_choice`** - Ensures invalid state values are rejected:
    * Uses Django's `full_clean()` validation.
    * Expects `ValidationError` for invalid choices.
* **`test_issue_github_comment_count_rejects_negative`** (updated) - Verifies negative values are rejected:
    * Previously documented that negative values were allowed.
    * Now properly tests that `PositiveIntegerField` rejects negative values.

### DIFF
--- a/website/migrations/0260_add_github_fields_to_issue.py
+++ b/website/migrations/0260_add_github_fields_to_issue.py
@@ -12,7 +12,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name="issue",
             name="github_comment_count",
-            field=models.IntegerField(default=0, help_text="Number of comments on the GitHub issue"),
+            field=models.PositiveIntegerField(default=0, help_text="Number of comments on the GitHub issue"),
         ),
         migrations.AddField(
             model_name="issue",
--- a/website/migrations/0260_add_github_fields_to_issue.py
+++ b/website/migrations/0260_add_github_fields_to_issue.py
@@ -1,0 +1,33 @@
+# Generated migration for adding GitHub integration fields to Issue model
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("website", "0259_add_domain_slug"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="issue",
+            name="github_comment_count",
+            field=models.IntegerField(default=0, help_text="Number of comments on the GitHub issue"),
+        ),
+        migrations.AddField(
+            model_name="issue",
+            name="github_state",
+            field=models.CharField(
+                blank=True,
+                choices=[("open", "Open"), ("closed", "Closed")],
+                help_text="Current state of the GitHub issue",
+                max_length=10,
+                null=True,
+            ),
+        ),
+        migrations.AddField(
+            model_name="issue",
+            name="github_fetch_status",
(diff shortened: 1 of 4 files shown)


website/migrations/0260_add_github_fields_to_issue.py
```diff
@@ -0,0 +1,33 @@
+# Generated migration for adding GitHub integration fields to Issue model
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("website", "0259_add_domain_slug"),
```
sidd190: Kindly address the comment here about non existent migration number, this is an issue which will throw error on applying migrations

sidd190 (changes requested):
### github_state field: remove null=True, use default="" (avoid NULL vs empty-string ambiguity)

Problem: In the migration and model, github_state is declared with null=True and blank=True. For CharFields Django best practice is to avoid null=True — prefer blank=True, default="". Using null=True creates two “empty” states (NULL and "") and complicates queries. CI also flagged this.

suggested Fixes:

Update migration’s AddField for github_state to remove null=True and set default="".
Update website/models.py Issue field definition to match.

Please take a look at this and the other comment. Looks good otherwise.

Pritz395 (changes requested):
Just need to update tests and you're all good. The model uses default="" for github_state (no null=True), but 5 tests still assert None:

Line 118: test_issue_created_with_default_github_fields
Line 187: test_issue_github_fetch_failed
Line 200: test_issue_update_github_fields
Line 290: test_multiple_issues_with_different_github_states
Line 303: test_issue_github_state_null_allowed

With default="", new issues have github_state="", not None. Update these assertions to assertEqual(issue.github_state, "") to match the model.

Pritz395 (changes requested):
- Migration: change IntegerField → PositiveIntegerField (line 15)
- Migration: add null=True to github_last_fetched_at field definition (line 39-42)

Optional:
- Remove unrelated template URL changes or note them in PR description
Clarify that fetching logic comes in a follow-up PR
- fix the migration field type mismatch before merging.

DonnieBLT:
is there a section where the data is added?  I only see the new fields and not the logic to add the data

PR #5146, opened 2025-12-02
Fix TransactionManagementError in IP restriction middleware

dRaniwal:
- Add try/except wrapper around _record_ip method to prevent IP logging failures from breaking requests
- Remove buggy 'if not transaction.get_autocommit()' check that always evaluated to True inside atomic block (autocommit is always False)
- Ensure duplicate record cleanup always runs within the atomic block

This fixes TransactionManagementError that occurs when exceptions (like Http404) are raised during request handling while a transaction is active.

## Summary by CodeRabbit

* **New Features**
  * IP restriction middleware now supports asynchronous request handling for improved responsiveness.

* **Bug Fixes**
  * Improved IP tracking with atomic updates to counts and agent data, and automated cleanup of duplicate records to maintain data consistency.

### DIFF
--- a/blt/middleware/ip_restrict.py
+++ b/blt/middleware/ip_restrict.py
@@ -143,29 +143,32 @@ class IPRestrictMiddleware:
         """
         Helper method to record IP information
         """
-        with transaction.atomic():
-            # create unique entry for every unique (ip,path) tuple
-            # if this tuple already exists, we just increment the count.
-            ip_records = IP.objects.select_for_update().filter(address=ip, path=path)
-            if ip_records.exists():
-                ip_record = ip_records.first()
+        try:
+            with transaction.atomic():
+                # create unique entry for every unique (ip,path) tuple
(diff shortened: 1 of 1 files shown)


dRaniwal:
> Should the tests be fixed by this? It seems they are still throwing an error

CI is still failing but it's testing against main, not my PR branch (due to pull_request_target workflow). The fix I pushed isn't being evaluated.

Can we either:

Merge the fix PR first so CI runs against the fixed code, or
Manually verify the fix works by running tests locally/in Docker?
The fix wraps [_record_ip](vscode-file://vscode-app/private/var/folders/7y/yt0v7y8d5jn5c289zwrl8ddh0000gn/T/AppTranslocation/0B8A8E88-3445-467A-836B-9A96D4E6157C/d/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) in try/except so transaction errors in the middleware don't break subsequent request handling.

PR #5187, opened 2025-12-05
feature: automatic detect and cover faces in images

### LINKED ISSUE #2527: Automatically detect and cover faces in images
We'll create a new function called `overlay_faces` that overlays a solid color on detected faces.

```python
import cv2
import numpy as np
from django.core.files.base import ContentFile
from django.shortcuts import render, redirect
from .forms import ImageUploadForm  # Assuming you have a form for image upload
from .models import UploadedImage  # Assuming you have a model to store uploaded images

def overlay_faces(image, color=(0, 0, 0)):
    # Convert image to grayscale
    gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)

    # Load OpenCV's pre-trained Haar Cascade face detection model
    face_cascade = cv2.CascadeClassifier(cv2.data.haarcascades + 'haarcascade_frontalface_default.xml')

    # Detect faces in the image
    faces = face_cascade.detectMultiScale(gray, scaleFactor=1.1, minNeighbors=5, minSize=(30, 30))

    # Overlay a solid color on each face found
    for (x, y, w, h) in faces:
        # Draw a filled rectangle over the face region with the specified color
        cv2.rectangle(image, (x, y), (x+w, y+h), color, thickness=cv2.FILLED)

    return image

def image_upload_view(request):
    if request.method == 'POST':
        form = ImageUploadForm(request.POST, request.FILES)
        if form.is_valid():
            image_instance = form.save(commit=False)

            # Read the uploaded image file
            image_file = request.FILES['image']
            np_img = np.frombuffer(image_file.read(), np.uint8)
            img = cv2.imdecode(np_img, cv2.IMREAD_COLOR)

            # Overlay faces in the image with a solid color (default is black)
            img_with_overlayed_faces = overlay_faces(img, color=(0, 0, 0))

            # Convert the image back to a format that can be saved in Django model
            _, buffer = cv2.imencode('.jpg', img_with_overlayed_faces)
            image_instance.image.save(image_file.name, ContentFile(buffer.tobytes()), save=False)

            # Save the image instance
            image_instance.save()
            return redirect('success_url')  # Replace 'success_url' with your actual success URL

    else:
        form = ImageUploadForm()
    return render(request, 'upload.html', {'form': form})
```

### Explanation of Changes

1. **`overlay_faces(image, color=(0, 0, 0))` Function**:
   - Similar to the previous `blur_faces` function, but instead of applying a blur, it draws a filled rectangle over each detected face.
   - The `color` parameter defines the color of the rectangle. The default is black `(0, 0, 0)` in BGR format.

2. **Usage**:
   - The `overlay_faces` function is called within the `image_upload_view` with a color of your choice to overlay the faces with that color.

### Customizing the Color

You can customize the `color` parameter when calling `overlay_faces`. The color should be in BGR format (Blue, Green, Red). For example:

- **Black**: `(0, 0, 0)`
- **White**: `(255, 255, 255)`
- **Red**: `(0, 0, 255)`
- **Green**: `(0, 255, 0)`
- **Blue**: `(255, 0, 0)`

### Example:

To overlay the faces with a solid red color, modify the call in `image_upload_view` like this:

```python
img_with_overlayed_faces = overlay_faces(img, color=(0, 0, 255))  # Red color
```

This setup will now overlay a solid color on any detected faces in the uploaded images.

swaparup36:
Resolves https://github.com/OWASP-BLT/BLT/issues/2527

on creating a bug report on /report when an image is uploaded that contain any human face, then those faces are detected and covered automatically before uploading.

These is the original image -
growth-marketing-1-20250617-8634429

Now this is the image after applying face detection and covering before uploading -
Screenshot 2025-12-06 021920

## Summary by CodeRabbit

* **New Features**
  * Uploaded bug-report screenshots can have faces automatically masked; availability is detected at upload time.
  * Tweet links now use the X domain format.

* **Bug Fixes / Reliability**
  * Image processing is optional with safe fallbacks so uploads won’t fail if processing isn't available.
  * Processed images are consistently saved with a JPG extension when applicable.

* **Chores**
  * Added OpenCV runtime dependency and system packages; installation now fails fast on errors.


website/utils.py
```diff
@@ -1129,3 +1129,222 @@ def get_default_bacon_score(model_name, is_security=False):
...
+        original_name = image_file.name
+        if "." in original_name:
+            name_without_ext = ".".join(original_name.split(".")[:-1])
+        else:
+            name_without_ext = original_name
+
+        new_filename = f"{name_without_ext}_processed.jpg"
+
+        # Create new UploadedFile with processed content
+        processed_file = ContentFile(buffer.tobytes(), name=new_filename)
```
sidd190: Consider maybe a `buffer_bytes = buffer.tobytes()` statement to cache it instead of calling it twice. Just a nitpick, not a blocker though.

Dockerfile
```diff
@@ -35,8 +40,11 @@ COPY pyproject.toml poetry.lock* ./
 # Clean any existing httpx installation and update pip
 RUN pip uninstall -y httpx || true
 RUN pip install --upgrade pip
-# Install dependencies with Poetry
-RUN poetry install --no-root --no-interaction
+# Install dependencies with Poetry - with retry and fallback
+RUN poetry install --no-root --no-interaction --no-ansi || \
```
sidd190: #### Poetry fallback is unsafe
The current fallback installs only four packages if poetry install fails and then continues building. This produces a “successfully built” image with a partially missing Python dependency set, leading to unpredictable runtime failures.

Consider this, we probably want to follow, fail early and loud rationale here. Maybe do something like -
either, no fallback, quick and loud fail. (Better, quicker and safer) - 
`RUN set -e; \
    poetry install --no-root --no-interaction --no-ansi || \
    (echo "ERROR: Poetry install failed. Please fix dependency configuration." >&2 && exit 1)`

OR something like

`RUN poetry install --no-root --no-interaction --no-ansi || \
    (echo "Poetry install failed, falling back to pip install $(poetry export --without-hashes --format=requirements.txt)" && \
     poetry export --without-hashes --format=requirements.txt | pip install -r /dev/stdin)`

sidd190 (changes requested):
Left a few comments. Please resolve the important ones

pyproject.toml
```diff
@@ -8,7 +8,7 @@ license = "AGPLv3"
 [tool.poetry.dependencies]
 python = ">=3.11.2,<3.14"
 python-dotenv = "^1.1.0"
-Django = "^5.2.9"
+Django = "^5.1.8"
```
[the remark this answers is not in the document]
swaparup36: Actually it was a mistake from my side, I'm really sorry for that. Thanks for pointing it out.

sidd190 (approved):
LGTM! It's good to go as long as you've tested it on more than 1 image on your part.

PR #5204, opened 2025-12-07
fix(search): Resolve search functionality and recent history issues

### LINKED ISSUE #5203: BUG (search bar) : No Results &  Recent Searches Not Appearing.

Nachiket-Roy:
### This PR fixes two critical bugs affecting the search functionality:

1. Search type filters weren't working - All searches returned results from all models regardless of selected type
2. Recent searches weren't appearing - Search history wasn't being saved properly
CLOSES : #5203 

## Summary by CodeRabbit

* **New Features**
  * Per-type search modes plus an “all” view with auth-aware visibility and safer wallet handling
  * Search history logging with deduplication, long-query truncation+hashing, and configurable per-user limits

* **Improvements**
  * Centralized result context population with enriched metadata, badges and URL enrichment
  * Lazy/on-demand result counts and consistent recent searches; home view adapts to auth state

* **Bug Fixes**
  * History logging and wallet retrieval errors are now handled without breaking searches

### DIFF
--- a/website/views/core.py
+++ b/website/views/core.py
@@ -649,6 +649,7 @@ def search(request, template="search.html"):
                 if d:
                     org.absolute_url = d.get_absolute_url()
             context = {
+                "request": request,
                 "query": query,
                 "type": stype,
                 "organizations": organizations,
@@ -656,6 +657,7 @@ def search(request, template="search.html"):
 
         elif stype == "projects":
             context = {
+                "request": request,
                 "query": query,
                 "type": stype,
                 "projects": Project.objects.filter(Q(name__icontains=query) | Q(description__icontains=query)),
@@ -663,6 +665,7 @@ def search(request, template="search.html"):
 
         elif stype == "repos":
             context = {
+                "request": request,
                 "query": query,
                 "type": stype,
                 "repos": Repo.objects.filter(Q(name__icontains=query) | Q(description__icontains=query)),
@@ -680,6 +683,7 @@ def search(request, template="search.html"):
                 if d:
                     org.absolute_url = d.get_absolute_url()
             context = {
+                "request": request,
                 "query": query,
                 "type": stype,
                 "tags": tags,
@@ -692,6 +696,7 @@ def search(request, template="search.html"):
 
         elif stype == "languages":
             context = {
+                "request": request,
                 "query": query,
                 "type": stype,
                 "repos": Repo.objects.filter(primary_language__icontains=query),
@@ -759,7 +764,7 @@ def search(request, template="search.html"):
                             result_count = len(items)
 
             except Exception as e:
-                logger.error("Error calculating result count")
+                logger.exception("Error calculating result count")
                 result_count = 0
 
             # Atomic operation: check for duplicates, create entry, and cleanup excess entries
--- a/website/views/core.py
+++ b/website/views/core.py
@@ -572,240 +572,234 @@ def search(request, template="search.html"):
     context = {}
 
     if query:
-        # Search across multiple models
-        organizations = Organization.objects.filter(name__icontains=query)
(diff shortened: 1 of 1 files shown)


arnavkirti (approved):
Thank you for taking up this issue. I was also facing this problem. Your implementation looks good to me. Thanks!!

sidd190 (commented):
Looks good overall, just a few things to confirm, test and address. Only the first is a blocker, others are just additive, nice to have nitpicks!

### 1. Hidden Issue Access
The current pattern:
```
.exclude(Q(is_hidden=True) & ~Q(user_id=request.user.id))
```
is not safe when `request.user` is anonymous (`id = None`). This can leak hidden issues for entries with no owner. Use an explicit authenticated/anonymous split in **every** Issue query:
```
if request.user.is_authenticated:
    qs = qs.exclude(Q(is_hidden=True) & ~Q(user_id=request.user.id))
else:
    qs = qs.exclude(is_hidden=True)
```
Apply this consistently to issues, labels, tags (`matching_issues`), and any other `Issue.objects.filter(...)` block.

### 2. Logging & PII
Logging currently uses both raw and `repr` formatting. Always use `!r` when logging truncated queries to avoid leaking raw input:
```
truncated_query[:50] if truncated_query else None!r
```
Also remove duplicated `truncated_query:` entries in the same log message.

### 3. Exception Handling
Remove any remaining bare:
```
except Exception:
```
Use either specific DB exceptions or `logger.exception("Unexpected error")` so that real failures aren’t swallowed silently.

### 4. Invalid stype
Add a fallback for unknown `stype` values, e.g.:
```
if stype not in allowed_types:
    stype = "all"
```
This prevents unpredictable behavior on malformed requests.

### 5. Context Consistency
Ensure every branch includes `"request": request` in the context. Some branches omit it while others include it. Consistency matters if templates depend on `request`.

### Tests Recommended
- Hidden issue access (authenticated vs anonymous)
- Search history 50-limit cleanup
- Logging does not expose raw query strings

If this is implemented/addressed, it's good to go.

Nachiket-Roy:
@coderabbitai Could you please fact-check the reviewer’s comments below?



**1. Hidden Issue Access**
The current pattern:

```python
.exclude(Q(is_hidden=True) & ~Q(user_id=request.user.id))
```

is not safe when `request.user` is anonymous (`id = None`). This can leak hidden issues for entries with no owner. An explicit authenticated/anonymous split should be used in every Issue query:

```python
if request.user.is_authenticated:
    qs = qs.exclude(Q(is_hidden=True) & ~Q(user_id=request.user.id))
else:
    qs = qs.exclude(is_hidden=True)
```

This should be applied consistently to issues, labels, tags (`matching_issues`), and any other `Issue.objects.filter(...)` blocks.

**2. Logging & PII**
Logging currently uses both raw and `repr` formatting. Always use `!r` when logging truncated queries to avoid leaking raw input:

```python
truncated_query[:50] if truncated_query else None!r
```

Also, remove duplicated `truncated_query:` entries within the same log message.

**3. Exception Handling**
Remove any remaining bare:

```python
except Exception:
```

Use either specific database exceptions or:

```python
logger.exception("Unexpected error")
```

so real failures are not swallowed silently.

**4. Invalid `stype`**
Add a fallback for unknown `stype` values, for example:

```python
if stype not in allowed_types:
    stype = "all"
```

This prevents unpredictable behavior on malformed requests.

**5. Context Consistency**
Ensure every branch includes `"request": request` in the context. Some branches omit it while others include it. Consistency is important if templates depend on `request`.

**Tests Recommended**

* Hidden issue access (authenticated vs anonymous)
* Search history 50-limit cleanup
* Logging does not expose raw query strings

Nachiket-Roy:
@coderabbitai could you tell me why the test failed?

PR #5274, opened 2025-12-12
feature [admin]: add superuser dashboard for user activity visibility

### LINKED ISSUE #3387: Django Management command for deleting unverified users

Ankit-singh-dot:
closes #3387 
This PR implements a superuser-only admin dashboard to view user activity,
as requested in the updated scope of Issue #3387.

Changes:
- Adds an Activity column in Django admin based on last_login
- Allows filtering users by Active / Inactive status
- Preserves existing admin functionality
- No automated deletion or background cleanup

Activity is intentionally defined using last_login only to avoid
assumptions about related models and keep the implementation safe
and extensible.

Screenshot 2025-12-13 at 3 52 05 AM

## Summary by CodeRabbit

* **New Features**
  * Bulk user deactivation action in the admin interface
  * Activity status filter for viewing active vs. inactive users
  * Visual activity indicators in the user list
  * Counts for user activity (bug reports, forum posts/comments) shown in the list
  * Admin user-management views and actions restricted to superusers


DonnieBLT (changes requested):
Activity metrics per user (e.g., bugs reported, comments, last login)

website/admin.py
```diff
@@ -1271,3 +1276,117 @@ class UserTaskSubmissionAdmin(admin.ModelAdmin):
...
+            ("active", "Active"),
+            ("inactive", "Inactive"),
+        )
+
+    def queryset(self, request, queryset):
+        if self.value() == "active":
+            return queryset.exclude(last_login__isnull=True)
+        if self.value() == "inactive":
+            return queryset.filter(last_login__isnull=True)
+        return queryset
```
DonnieBLT: have this also take into account the other activity

DonnieBLT (dismissed):
please see comment and please simplify it all

Nachiket-Roy (dismissed):
@coderabbitai has verified a few points. Please address them.

DonnieBLT:
Are you still interested in working on this?

yashhzd (commented):
Good idea — having activity visibility in the admin panel is useful for understanding engagement. The annotated queryset approach with `Count()` is the right call to avoid N+1 queries when displaying counts in the list view.

A few observations:

**Duplicate User admin registration**
The file currently has `admin.site.register(User, UserAdmin)` at line ~957 and then unregisters+re-registers with `CustomUserAdmin` at the bottom. This means the first registration is dead code. Removing the old `register(User, UserAdmin)` call (and the unguarded `admin.site.unregister(User)` earlier) would clean things up — which is what DonnieBLT seems to be asking for when he says "simplify it all."

**User import shadowing**
Both `from django.contrib.auth.models import User` and `User = get_user_model()` exist in the same file. Only one is needed. Since `get_user_model()` is the recommended Django pattern (works with custom user models), the direct import can be removed.

**Deactivate count accuracy**
`queryset.update(is_active=False)` reports the total row count, which includes users that were already inactive. Filtering to `.filter(is_active=True)` first would give an accurate "Deactivated N users" message.

**Simplification per maintainer feedback**
DonnieBLT's last review asked to simplify. The current diff is 139 lines. You could tighten it by removing the duplicate imports/registrations and inlining the simpler permission checks. The core value (activity columns + filter + deactivate action) is solid — the noise is mostly in the import/registration plumbing.

The superuser permission gating on `has_module_permission`, `has_view_permission`, `has_change_permission`, and `get_actions` looks correct and thorough.

DonnieBLT (changes requested):
This is still in draft mode. Please take it out of draft mode when it's ready.

PR #5317, opened 2025-12-16
Fix PostgreSQL ValueError by loading env variables in manage.py

armanraymagit:
I Identified this Issue during my local setup using python env and docker.

This PR fixes this Value Error related to the PostgreSQL port when running` python manage.py` during local setup.

- Added `load_dotenv()` in manage.py to load environment variables.
- Only affects local setup; no functional changes to the application.

Without loading environment variables, the PostgreSQL port from .env is not recognized, Causing a Value Error.
Also did `python manage.py test`
Screenshot 2025-12-16 143456

## Summary by CodeRabbit

* **Chores**
  * Refactored application startup and initialization sequence to streamline environment loading and entry-point behavior.
  * Internal startup logic reorganized for clearer execution flow.
* **User impact**
  * No user-facing changes; behavior and functionality remain the same.


DonnieBLT (changes requested):
are these not getting set in the settings?  What does the run_if_local do? @coderabbit feel free to answer too

PR #5387, opened 2026-01-02
Remove unused CaptchaField from Issue model to fix schema mismatch

### LINKED ISSUE #5386: Issue model defines CaptchaField without corresponding database migration
Problem
The Issue model defines a CaptchaField, but there is no database migration that adds this field to the database schema. As a result, the model and database are currently out of sync.

This was discovered while adding test coverage: Issue.objects.create() succeeds without providing a captcha value because the column does not actually exist in the database.

Location
Model: website/models.py (Issue model)
Field: CaptchaField
Issue: No migration exists to create the corresponding database column.

Why this matters
- Creates model-schema inconsistency
- Can cause unexpected runtime errors in the future
- Makes migrations and refactors harder
- Tests may give a false sense of correctness

Suggested next steps
One of the following should be done:
1. Add a migration to properly create the CaptchaField in the database, if the field is intended to be used.
2. Remove the field from the model if it is deprecated or no longer needed.

Nachiket-Roy:
This PR removes the unused CaptchaField from the Issue model. The field was defined in the model but never added to the database schema via migration, resulting in a model-database mismatch.

Since the field is no longer used anywhere in the codebase, removing it resolves the inconsistency and avoids potential future issues.

Problem Addressed
- CaptchaField existed in website/models.py
- No migration ever created the corresponding database column
- Led to silent model-schema divergence
- Could cause confusion, migration conflicts, or runtime errors later

Why removal is the right fix
- The field is unused and unreferenced
- Avoids introducing unnecessary migrations for dead code
- Keeps the data model clean and accurate
- Aligns model definitions with the actual database schema

Scope
- Removes the CaptchaField from the Issue model
- No functional or behavioral changes elsewhere
- No impact on existing data or migrations

Testing
- Existing tests pass
- No new tests required, as this change removes unused code only
closes: #5386 

## Summary by CodeRabbit

* **Refactor**
  * Removed CAPTCHA requirement from issue submission form.


Jayant2908 (approved):
LGTM!

S3DFX-CYBER (approved):
Great catch on the schema mismatch, @Nachiket-Roy. Removing dead code definitely keeps the migrations cleaner.

One quick security question: Since we are removing the CaptchaField from the model, have we verified that the frontend form for Issue submission still has adequate rate-limiting or alternative protection against bot-spam? I want to make sure we aren't losing a layer of defense here.

Nachiket-Roy:
@S3DFX-CYBER not sure what you are implying but this removed code is not in use anywhere https://github.com/OWASP-BLT/BLT/issues/5386#issuecomment-3705714508

Nachiket-Roy:
@S3DFX-CYBER, the CAPTCHA referenced in the UI exists only at the view layer and was never enforced or validated at the model or database layer. Since the backend does not consume or verify this field, removing it does not disable an active security control.
That said, I agree that if the intent is to fully enable CAPTCHA enforcement end-to-end, adding the missing migration and backend validation would be the correct approach.

Nachiket-Roy:
@coderabbitai, should CAPTCHA be fully enforced end-to-end (UI + backend) for these endpoints? Please consider the existing server-side protections and the trade-offs (complexity, dependency, accessibility) when evaluating whether this should be treated as a fix or a feature.

Nachiket-Roy:
@S3DFX-CYBER, coderabbit reviewed the full flow and confirmed that CAPTCHA enforcement is already handled at the view/form layer and that the model field was never part of the active implementation. From that perspective, removing it avoids dead schema and doesn’t reduce security. Happy to discuss further here, or we can also continue the conversation on Slack and loop in maintainers if you think that would be helpful.

DonnieBLT (changes requested):
do we need a migration for this?

PR #5406, opened 2026-01-07
fix: resolve CaptchaField model bug 

S3DFX-CYBER:
Description
This PR fixes a critical architectural bug where a form-level field was improperly placed in the database model, while also restoring core models to maintain system integrity and pass existing tests.
Key Changes
 * Database Schema Fix: Removed CaptchaField from the Issue model in website/models.py. As identified by Sentry, CaptchaField is a form-specific field and its presence in the model causes database migration and deployment failures.
 * Form Layer Validation: Re-implemented the captcha field within IssueForm in website/forms.py. This ensures bot protection remains active at the validation layer without requiring a corresponding database column.
 * Model Restoration: Restored core models including Job, Trademark, Organization, and others that were inadvertently removed. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed CAPTCHA validation from issue submissions.


Nachiket-Roy (changes requested):
@S3DFX-CYBER, why are you removing so many models? This doesn’t seem necessary, and it will likely cause the existing tests to fail.

Jayant2908 (commented):
@S3DFX-CYBER Isn't this a duplicate PR of PR#5387 by Nachiket?

S3DFX-CYBER:
Everything seems to be fine @DonnieBLT 
You can go further with your decision

Nachiket-Roy:
> If we align these, the final outcome should be:
> no dead model fields
> no migrations needed
> consistent Captcha validation on Issue/Signup forms

I’m not sure what you’re trying to convey here. This PR also removes the same dead code already addressed in my PR. Regardless of whether those specific lines are removed or retained, no migration would apply, since the field was never part of the active enforcement path.

CAPTCHA validation already exists at the view layer for both issue submission and signup, so the final state you describe is already achieved without introducing a migration.

DonnieBLT (changes requested):
if we are removing this from the model we'd also need a migration.  I will check the other PR

S3DFX-CYBER:
@DonnieBLT, I agree with you. To ensure Django's internal state remains consistent and to satisfy the maintainer requirements, I will include a migration that reflects the removal of this field from the Issue model.

PR #5508, opened 2026-01-21
Fix inconsistent Page Stats panel behavior across desktop and mobile

### LINKED ISSUE #5501: Page Stats container opens unintentionally and behaves inconsistently across devices
The Page Stats container currently opens unintentionally due to an overly broad interaction area and conflicting UI states. This results in accidental panel opening, inconsistent closing behavior, and a confusing user experience on both desktop and mobile devices.

### 1. Unintentional opening near the icon (Desktop & Wide Screens)
- On desktop and large screens, dragging the mouse **near the left or right side of the stats icon** causes the container to open.
- The container opens even when the user does **not** directly interact with the icon.
- This creates a mismatch between the visible interactive element (icon) and the actual trigger area.

---

### 2. Accidental opening on mobile devices
- On small screen devices, tapping or interacting **near** the stats icon can cause the container to open unintentionally.
- Users may click elsewhere on the screen, but the stats panel opens unexpectedly.
- This behavior is especially disruptive on mobile devices where screen space is limited.

---

### 3. Panel does not close when clicking outside (Desktop & Mobile)
- Once the stats container is opened:
  - Clicking anywhere else on the page does **not** close it.
  - Users must explicitly interact with the icon again to dismiss it.
- This behavior feels unintuitive on both desktop and mobile devices.

https://blt.owasp.org/issue/5805 found by codeXanuj at url: http://blt.owasp.org

codeXanu:
### Summary
This PR fixes multiple UX issues with the Page Stats panel, where the container was opening unintentionally, **like hovering ( clicking in mobile ) to the left or right side of the icon ( in the width of the panel )**, closing unexpectedly, and behaving inconsistently across desktop and mobile devices.

The interaction logic has been refined to clearly distinguish between **hover-based preview behavior** and **click-based intentional behavior**, resulting in a more predictable and user-friendly experience.

---

### Problems Addressed

- The Page Stats panel could open when the cursor was merely near the icon, rather than directly interacting with it.
- On desktop, the panel could close unexpectedly when the mouse moved out of the container, even after an intentional click.
- On mobile devices, clicking to the left or right of the icon opens the panel, and closing the panel by clicking outside does not work on desktop and mobile devices.
- Hover and click interactions were previously treated the same, leading to conflicting behavior.

---

### What Changed

- Introduced a clear distinction between **hover-opened (temporary preview)** and **click-opened (persistent)** panel states.
- Hovering over the stats icon now opens the panel as a preview.
- Clicking the icon while hovered converts the panel into a persistent, click-opened state instead of closing it.
- The panel now:
  - Closes on mouse leave only if it was opened via hover.
  - Remains open on mouse leave if it was opened via click.
  - Closes reliably when clicking outside or clicking the icon again.
- Mobile behavior is now consistent, opens only on clicking the icon, and closes on clicking the icon or outside the panel.

---

### Resulting Behavior

- Hover on icon → panel opens (preview)
- Mouse leaves container (hover-open) → panel closes
- Click on icon → panel opens and stays open
- Mouse leaves container (click-open) → panel remains open
- Click outside panel → panel closes
- Click icon again → panel closes
- Mobile interactions now work without requiring repeated taps

---

### Why This Approach

- Prevents accidental activation while preserving quick hover previews on desktop.
- Ensures intentional user actions (clicks) are respected.
- Avoids hover-only or proximity-based logic that causes unpredictable behavior.
- Keeps changes scoped and non-invasive to existing layout and animations.

---

### The changes could be seen in these videos:

https://github.com/user-attachments/assets/60e28804-52ae-442a-bd1d-f0f7d7984b01

https://github.com/user-attachments/assets/0b1b8651-3703-4961-8b18-1d72091a0c77

### Related Issue
Closes **#5501**

## Summary by CodeRabbit

* **New Features**
  * Page stats panel now supports explicit open/close modes with hover and click behaviors, including persistent open via click.
  * Click-outside closes the panel when in click-open mode.
* **Accessibility**
  * Toggle control updated with ARIA attributes and improved keyboard/mouse interaction for better screen-reader support.

### DIFF
--- a/website/templates/includes/page_stats.html
+++ b/website/templates/includes/page_stats.html
@@ -7,7 +7,10 @@
 <div id="pageStatsContainer"
      class="fixed bottom-0 left-4 z-50 transition-all duration-300 transform translate-y-[calc(100%-40px)] group">
     <!-- Chart Icon and Handle here -->
-    <div class="flex justify-center items-center h-10 w-16 mx-auto bg-white dark:bg-gray-800 rounded-t-lg shadow-md cursor-pointer">
+    <button type="button"
+            class="flex justify-center items-center h-10 w-16 mx-auto bg-white dark:bg-gray-800 rounded-t-lg shadow-md cursor-pointer"
+            aria-expanded="false"
+            aria-label="Toggle page statistics">
         <svg xmlns="http://www.w3.org/2000/svg"
              class="h-6 w-6 text-[#e74c3c]"
              fill="none"
@@ -15,7 +18,7 @@
              stroke="currentColor">
             <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M9 19v-6a2 2 0 00-2-2H5a2 2 0 00-2 2v6a2 2 0 002 2h2a2 2 0 002-2zm0 0V9a2 2 0 012-2h2a2 2 0 012 2v10m-6 0a2 2 0 002 2h2a2 2 0 002-2m0 0V5a2 2 0 012-2h2a2 2 0 012 2v14a2 2 0 01-2 2h-2a2 2 0 01-2-2z" />
         </svg>
-    </div>
+    </button>
     <!-- Stats Content -->
     <div class="bg-white dark:bg-gray-800 shadow-lg rounded-lg p-3 w-72">
         <div class="text-sm font-medium text-gray-700 dark:text-gray-300 mb-2">Page Views (Last 30 Days)</div>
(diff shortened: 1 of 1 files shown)


S3DFX-CYBER (commented):
One small change @codeXanu , Inside initializeChart, you have a setTimeout loop checking for the Chart global. While this works, if the library fails to load, this loop runs indefinitely. Adding a maxAttempts counter or an else log would be a bit safer for production debugging.

Jayant2908 (changes requested):
As @Nachiket-Roy Suggested the issue 1 is critical, other than do check out the issues mentioned by coderabbitai above.

S3DFX-CYBER (approved):
LGTM @codeXanu 
But I would recommend some suggestions they might not be necessary 

Since you've added great keyboard support by using a <button>, adding a listener for the Escape key to close the panel when openMode === 'click' would be the "cherry on top" for accessibility.

The setTimeout loop for Chart.js is functional, but if the script fails to load (e.g., ad-blocker or network error), it will loop forever. Consider adding a counter to stop trying after 10–20 attempts.

PR #5511, opened 2026-01-21
fix: Skip IP recording during tests to avoid transaction errors

Pritz395:
## Summary

This PR fixes intermittent `TransactionManagementError` issues during tests by hardening the `IPRestrictMiddleware` so it behaves safely in test environments while preserving visibility in production.

When tests use `transaction.atomic()` (e.g., for CVE-related logic), teardown can leave the transaction in a broken state. The IP restriction middleware then tries to record IPs inside its own `transaction.atomic()` block and can raise `TransactionManagementError`, causing tests and CI (including docker-test) to fail unexpectedly.

## Changes

- Update `IPRestrictMiddleware._record_ip` to:
  - Skip IP recording entirely when `settings.IS_TEST` or `settings.TESTING` is true.
  - Check `transaction.get_rollback()` and bail out early when the current transaction is already marked for rollback.
- Add a dedicated `except TransactionManagementError` handler that:
  - Logs at `DEBUG` in test environments to avoid noisy teardown failures.
  - Logs at `WARNING` with IP and exception details (and `exc_info=True`) in non-test environments so real production issues remain visible.
- Keep a generic `except Exception as e` handler to log any unexpected errors with full stacktrace.

## Rationale

- Keeps production behavior intact: IP recording and blocking still work as before, but transaction errors are clearly surfaced in logs.
- Makes test and CI runs more reliable: the middleware no longer crashes tests when a previous `transaction.atomic()` has already put the transaction into a rollback-only state.
- Cleanly separates this test/infra-oriented change from CVE feature work handled in PR #5057.

## Impact

- No database schema changes.
- No public API changes.
- Behavior change is limited to how the middleware records IPs when a transaction is already broken, and how those errors are logged in test vs non-test environments.

## Summary by CodeRabbit

* **Chores**
  * Improved middleware robustness: refined IP recording behavior and enhanced error logging for transaction-related failures to capture full context.
* **Tests**
  * Reduced noise and improved reliability by skipping IP recording during test runs and adjusting logging to avoid spurious errors.


Pritz395:
@coderabbitai give the fixes for all pending issues.

Jayant2908 (changes requested):
Great fixing this error as it occurs a lot of times. Although I found two issues with the approach, not sure If I am correct Can you please look into the above review by coderabbitai!?

blt/middleware/ip_restrict.py
```diff
@@ -138,6 +140,10 @@ def _record_ip(self, ip, agent, path):
         """
         Helper method to record IP information
         """
+        # Skip IP recording during tests to avoid transaction management issues
+        if getattr(settings, "IS_TEST", False) or getattr(settings, "TESTING", False):
```
S3DFX-CYBER: I noticed that while you hardened _record_ip with the specific TransactionManagementError block, the increment_block_count method  still uses a generic Exception catch.

Suggestion: Update increment_block_count to use the same if IS_TEST logic for its error handling so the behavior is consistent across the whole middleware.

Pritz395:
@coderabbitai what's causing the test fail?

PR #5627, opened 2026-02-09
chore: Removed Available mentors

Jayant2908:
Removed the "Available Mentors" since each mentor already has one mentee, and Donnie mentioned that only one mentee is allowed per mentor.

Screenshots-
image

## Summary by CodeRabbit

* **Chores**
  * Removed the "Available Mentors for Future Students" section from the GSOC page, including the mentor cards and their descriptions.
  * Page now flows directly into the Past Mentors section with no other structural changes.
  * Change is small and low review effort.

### DIFF
--- a/website/templates/gsoc.html
+++ b/website/templates/gsoc.html
@@ -1031,46 +1031,6 @@
                         </div>
                     </div>
                 </div>
-                <!-- Available Mentors Section -->
-                <div class="mt-16 bg-gray-50 dark:bg-gray-900 rounded-xl p-8 border border-gray-200 dark:border-gray-700">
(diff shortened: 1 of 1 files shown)


Jayant2908:
@Nachiket-Roy I'll add the available mentors block back but I dont see any duplication.

Jayant2908:
I have updated the description with a screenshot, Do check it out.

Nachiket-Roy:
> @Nachiket-Roy I think as its empty right now, we can remove it. If any mentor comes, he can again add it. That way its better ig

 Comment out the code rather than removing?

S3DFX-CYBER (approved):
Ok then LGTM

PR #5634, opened 2026-02-10
bug-fix: markdown errors

### LINKED ISSUE #5633: bug fix: fix the markdown - link, image , escape chars in reporting bugs.
Image

e-esakman:
### After
image

### changes
- Fixed bold, italic, and link rendering in markdown
- Escaped markdown input before rendering

## Summary by CodeRabbit

* **New Features**
  * Issue references (#123) are automatically converted to clickable links in descriptions.
  * Markdown rendering supports embedded images and richer formatting.

* **Improvements**
  * Confirmation warning for issue deletion emphasizes irreversible action.
  * External links include safer attributes (noopener noreferrer).
  * UI refinements and dark‑mode styling for better readability, including tooltip and control styling.
  * Resolve/delete actions provide alert feedback and refresh on success.


yashhzd (commented):
## Peer Review: PR #5634 — "bug-fix: markdown errors"

Good improvements to markdown rendering, dark mode, and security (`rel="noopener noreferrer"`, SRI hashes). However, there are several issues — some of which are functional bugs.

---

### Critical

**1. `getCookie()` has broken indentation — potential runtime issue**
The `for` loop and inner `if` block are de-indented out of the parent `if (document.cookie ...)` scope. The closing braces are misaligned, making the parse tree ambiguous. Both `deleteIssue` and `resolveIssue` depend on this function for CSRF tokens — if it breaks, all destructive actions silently fail.

```javascript
if (document.cookie && document.cookie !== '') {
    const cookies = document.cookie.split(';');
for (let i = 0; i < cookies.length; i++) {   // ← wrong indentation
    const cookie = cookies[i].trim();
if (cookie.substring(0, name.length + 1) === (name + '=')) {  // ← wrong
```

**2. `ALLOWED_URI_REGEXP` has a bug — uses single slash instead of double**
```javascript
ALLOWED_URI_REGEXP: /^(?:(?:https?|mailto):\/)/i
```
This matches `https:/` (one slash) but NOT `https://` (two slashes). This means DOMPurify will **strip the `href` from every markdown link**, rendering all links as plain underlined text with no destination. Should be `/^(?:(?:https?|mailto):\/\/)/i`.

**3. `img` + `src` allowed in DOMPurify without HTTPS-only restriction**
Adding `img` to `ALLOWED_TAGS` and `src` to `ALLOWED_ATTR` allows tracking pixels from any `http://` or `https://` domain. For a security-focused platform, user-supplied markdown can now embed external image requests that leak visitor IP addresses.

---

### Medium

**4. `console.log("Script loaded")` left in production**
Debug logging at line ~510 is committed to main. Every visitor sees this in their browser console.

**5. `deleteIssue` `.catch()` silently swallows errors from the user's perspective**
The `.catch()` only calls `console.error()` — the user gets no feedback on network failures (timeout, offline). The `else` branch has `alert()` but `.catch()` does not. This is a regression from the original `$.notify()` that showed feedback in both cases.

**6. `mode: 'same-origin'` replaced `credentials: 'same-origin'` in `deleteIssue`**
These are different fetch options. `credentials` controls whether cookies are sent; `mode` controls CORS behavior. The CSRF token is sent via header (from `getCookie`), but the session cookie may not be sent, which could break server-side `request.user` authentication checks.

**7. `{% url 'resolve' object.id %}` inside `<script>` without `|escapejs`**
Currently safe since `object.id` is numeric, but if the URL pattern ever changes to include strings, this becomes a template injection vector. Use `|escapejs` for consistency and safety.

---

### Minor

**8. Edit button silently removed**
The `id="edit-issue"` button was removed without mention in the PR description. If it was dead UI, that should be documented. If it was functional, this is a feature regression.

**9. `const markdownContent = decodedContent` is pointless indirection**
`markdownContent` is assigned from `decodedContent` and used once. Remove the intermediate variable.

**10. Manual Unicode escape decoding is fragile**
The regex `/\\u([\d\w]{4})/gi` doesn't handle surrogate pairs (emoji, characters above U+FFFF) or `\xNN` hex escapes. Descriptions with emoji may be silently corrupted.

**11. `data-markdown` fallback is now dead code**
The `|| bugReportElement.textContent` fallback path always yields an empty string since the comment was moved outside the `<p>` tag, making the `<p>` body empty.

**12. Trailing whitespace on `window.location.href = '/'; `**
Minor lint issue — trailing space after semicolon.

---

### Nit

- Inconsistent indentation throughout new `<script>` blocks (mix of 2-space and 4-space)
- SRI hash algorithm inconsistency: `sha384-` for markdown-it vs `sha256-` for DOMPurify
- `rel="noopener noreferrer"` added to one external link but not to other external links in the sidebar include — partial security fix gives false impression of completeness

---

The ALLOWED_URI_REGEXP bug (item #2) is likely breaking all markdown links in production right now and should be fixed urgently. The `getCookie` indentation (item #1) should also be verified and cleaned up.

PR #5661, opened 2026-02-12
Batch-fetch issues in hunt verification to fix N+1 queries

yashhzd:
## Summary\n\n- Replace O(N) `Issue.objects.get(pk=issue_id)` calls inside POST loop with single batch query\n- Use `bulk_update()` to persist all changes in one query instead of `issue.save()` per iteration\n- Replace unsafe `request.POST[\"key\"]` with `.get()` to prevent KeyError on malformed requests\n- Move `checkAll` lookup outside the loop (was re-checked via try/except on every iteration)\n\n## Problem\n\nIn `organization_hunt_results`, when processing hunt verification form submissions, the code previously:\n1. Called `Issue.objects.get(pk=issue_id)` for **every POST parameter** (N+1 query pattern)\n2. Called `issue.save()` individually for each issue (N write queries)\n3. Used `request.POST[\"checkAll\"]` inside a try/except **on every loop iteration**\n4. Used unsafe `request.POST[\"submit\"]` that raises KeyError if missing\n\n## Solution\n\n- Collect all issue IDs upfront, batch-fetch with `filter(pk__in=..., hunt=hunt)`\n- Build a map for O(1) lookups during iteration\n- Use `bulk_update()` for a single write query at the end\n- Check `checkAll` once before the loop\n- Use `.get()` for safe POST key access\n\n## Test Plan\n\n- [ ] Test hunt verification form with multiple issues\n- [ ] Test \"Check All\" functionality\n- [ ] Test save and publish actions\n- [ ] Verify no regressions in hunt results display


yashhzd:
Closing — consolidating into fewer, higher-impact PRs.

PR #5662, opened 2026-02-12
Fix unclosed file handle in IssueCreate screenshot save

yashhzd:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved screenshot handling: normalized filename behavior, robustly detects missing screenshots and shows a clear error while returning the report form.
  * Ensures uploaded screenshot files are properly closed after processing to prevent resource leaks.

* **Style**
  * Minor formatting tweaks that do not change behavior.


gojo-satorou-v7 (changes requested):
definitely a needed fix PR. but please run the pre-commit.

gojo-satorou-v7 (approved):

yashhzd:
Addressed CodeRabbit's review in commits 1077b59 and 0fefc3d:

**Double file extension bug (fixed):** Removed the unnecessary extension extraction in the `screenshot-hash` section (around line 714). The hash already gets `.png` appended on the next line, so preserving the original extension was creating filenames like `screenshot-uuid.jpg.png`.

**Indentation concern (no issue):** Verified the indentation is correct — lines 718-725 are inside the `if` block with 12-space indent (one level deeper than the 8-space `if` at line 714). The blank line at 717 doesn't end Python blocks.

Note: The file upload section (line 1411-1416) intentionally keeps the extension extraction since uploaded files need their original extension preserved — that's a different code path from the screenshot-hash handling.

yashhzd:
All the CodeRabbit suggestions from the latest review are already addressed in the current code:

1. **Double extension fix** — `screenshot-hash` is now set to `filename[:99] + str(uuid.uuid4())` without appending the original extension, so when `.png` is added later there's no double extension
2. **NoneType guard** — the `screenshot_name` block is inside the `if self.request.POST.get("screenshot-hash"):` guard
3. **FileNotFoundError handling** — both the create flow (`form_valid`) and update flow already wrap `default_storage.open()` in `try/except FileNotFoundError` with user-friendly error messages and `try/finally` for proper file handle cleanup

The `changes-requested` may be from an intermediate commit that's since been updated.

PR #5666, opened 2026-02-12
Eliminate N+1 queries in get_score API endpoint

yashhzd:
## Summary\n\n- Eliminate 5 per-user database queries in the `get_score` API endpoint\n- Use `select_related('userprofile')` to fetch profiles in the same query\n- Use the already-annotated `total_score` instead of re-aggregating\n\n## Problem\n\nThe `get_score()` endpoint builds a leaderboard by iterating over all users with scores. For each user, it ran **5 separate queries**:\n\n1. `Points.objects.filter(user=each.id).aggregate(total_score=Sum('score'))` - redundant, already annotated\n2. `UserProfile.objects.filter(user=each.id).values('user_avatar')` - avatar\n3. `UserProfile.objects.filter(user=each.id).values('title')` - title\n4. `UserProfile.objects.filter(user=each.id).values('follows')` - follows\n5. `UserProfile.objects.filter(user=each.id).values('issue_saved')` - saved issues\n\nWith 100 users on the leaderboard, this meant **501 queries** instead of 1.\n\n## Solution\n\n- Use `select_related('userprofile')` to JOIN the profile in the initial query\n- Access `user.total_score` directly (already annotated)\n- Access `user.userprofile` attributes directly (prefetched via JOIN)\n- Reduces total queries from **5N + 1** to **1** (plus 2N for M2M counts)\n\n## Test Plan\n\n- [ ] Verify `/api/v1/score/` returns the same JSON structure\n- [ ] Check leaderboard page renders correctly\n- [ ] Run existing test suite for regressions

### DIFF
--- a/website/views/user.py
+++ b/website/views/user.py
@@ -214,7 +214,7 @@ def profile_edit(request):
 
                 messages.info(
                     request,
-                    "A verification link has been sent to your new email. " "Please verify to complete the update.",
+                    "A verification link has been sent to your new email. Please verify to complete the update.",
                 )
                 return redirect("profile", slug=request.user.username)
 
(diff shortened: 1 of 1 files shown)


yashhzd:
Fixed the `AutoOneToOneField` auto-creation issue. Replaced `select_related("userprofile")` with a separate batch query using `UserProfile.objects.filter(user__in=temp_users)` and a dictionary lookup. This avoids triggering auto-creation on the GET endpoint while still keeping the optimization at O(2) queries instead of O(N+1).

PR #5670, opened 2026-02-12
fix: use get_object_or_404 and require POST for destructive team actions

yashhzd:
## Summary\n\n- Replace bare `Organization.objects.get()` in `join_requests` with `get_object_or_404` to return proper 404 instead of 500 on invalid team ID\n- Add `@require_POST` decorator to `delete_team` and `leave_team` views to prevent state-changing operations via GET requests (CSRF protection)\n- Remove redundant `request.user.is_authenticated` checks inside `@login_required` views where the decorator already guarantees authentication\n\n## Changes\n\n**`website/views/teams.py`**:\n- `join_requests`: `Organization.objects.get()` → `get_object_or_404()` — prevents uncaught `DoesNotExist` exceptions (500 errors) when a team ID is invalid or not found\n- `delete_team`: Added `@require_POST` — prevents team deletion via GET requests, which could be exploited via CSRF or prefetch attacks\n- `leave_team`: Added `@require_POST` — same protection as above for leaving a team\n- Both `delete_team` and `leave_team`: Removed redundant `if request.user.is_authenticated` guard — `@login_required` already guarantees the user is authenticated\n\n## Test plan\n\n- [ ] Verify `join_requests` returns 404 for invalid team IDs instead of 500\n- [ ] Verify `delete_team` rejects GET requests with 405\n- [ ] Verify `leave_team` rejects GET requests with 405\n- [ ] Verify all three views still work correctly for valid POST requests

### DIFF
--- a/website/views/teams.py
+++ b/website/views/teams.py
@@ -8,9 +8,10 @@ from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from django.db import IntegrityError
 from django.db.models import Count
 from django.http import JsonResponse
-from django.shortcuts import redirect, render
+from django.shortcuts import get_object_or_404, redirect, render
 
 # Create your views here.
+from django.views.decorators.http import require_POST
 from django.views.generic import TemplateView
 from rest_framework.response import Response
 from rest_framework.views import APIView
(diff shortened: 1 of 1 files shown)


sidd190 (approved):
LGTM! It feels I'm technically just reviewing same thing over and over. Just make sure to try and merge these fixes somehow. 

Thanks for the PRs. Nice recon.

PR #5672, opened 2026-02-12
fix: use get_object_or_404 and add @require_POST in education views

yashhzd:
## Summary by CodeRabbit

* **Bug Fixes**
  * Create vs update now return correct success status codes (201 for create, 200 for update); missing courses show proper errors.
  * Permission checks now prevent unauthorized course edits and redirect on failure.

* **Improvements**
  * Course create/update requires authentication, validates required fields and thumbnail type, and enforces edit permissions; returns course ID on success.
  * Lecture deletion requires a submitted request, includes the lecture title in success messages, reorders remaining lectures, and redirects appropriately.
  * Marking lectures complete now accepts POST requests with JSON and returns updated progress.


yashhzd:
Fixed both CodeRabbit suggestions:

1. Moved `user_profile = get_object_or_404(UserProfile, user=request.user)` before the `try` block so the `Http404` actually propagates instead of being swallowed by the `except Exception` handler (Option B from the review)
2. Removed the redundant `if request.method == "POST"` check and dead `else` branch since `@require_POST` already handles method enforcement

Both issues were valid catches - thanks for flagging them.

yashhzd:
Addressed latest CodeRabbit feedback:

- **Level validation**: Added validation against `Course.LEVEL_CHOICES` — returns 400 for invalid values
- **`request.user.userprofile`**: Switched from `get_object_or_404(UserProfile, ...)` to `request.user.userprofile` for consistency with `AutoOneToOneField` pattern used by other views
- **`bulk_update`**: Lecture reordering in `delete_lecture` now uses `bulk_update` instead of per-row saves
- **Pre-commit**: Ran ruff — also auto-removed unused `UserProfile` import

yashhzd:
Regarding the two actionable comments on the latest commit:

`UserProfile` uses `AutoOneToOneField` (from `django-annoying`, defined at line 903 in `models.py`), which auto-creates the profile on first access. So `request.user.userprofile` never raises `RelatedObjectDoesNotExist` — it's safe without `get_object_or_404`.

This is also consistent with the rest of the file — lines 54, 73, 125, 184, 286, and 497 all use `request.user.userprofile` directly without any error handling for the same reason.

So no changes needed here — the current code is correct.

DonnieBLT (dismissed):
Please get a peer review.

DonnieBLT:
moving to BLT-University

PR #5678, opened 2026-02-12
fix(update_user_status): fix AttributeError and use bulk update

yashhzd:
## Summary by CodeRabbit

* **Refactor**
  * Improved bulk inactive-user deactivation to run as a single, efficient update rather than per-user operations.
  * Adjusted inactivity detection logic and threshold to more consistently identify dormant accounts.
  * Reduced database load and execution time, and changed command output to a concise summary showing the total number of accounts deactivated.


yashhzd:
Addressed the CodeRabbit feedback - pushed a fix that handles users with NULL `last_login` by using `Q` objects:

```python
Q(last_login__lt=inactive_threshold)
| Q(last_login__isnull=True, date_joined__lt=inactive_threshold)
```

This also deactivates users who registered but never logged in if their `date_joined` is older than the threshold. Also switched to `self.style.SUCCESS` for colored output.

PR #5679, opened 2026-02-12
fix(views): guard int() on pagination params against ValueError

yashhzd:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of pagination parameters to gracefully default to valid values when invalid inputs are provided, preventing errors when accessing pages with malformed requests.


Jayant2908 (approved):
Looks good!

PR #5681, opened 2026-02-12
fix(education): safe int parsing and bulk section reorder

yashhzd:
## Summary\n\n### Input Validation\n- `add_section` and `add_lecture`: Wrap `int(request.POST.get(\"order\", 0))` in try/except to prevent `ValueError` crash when non-numeric input is submitted.\n\n### Performance\n- `delete_section`: Replace per-section `.save()` loop during reorder with `bulk_update()`, reducing N database writes to a single query.\n\n## Test Plan\n- [ ] Adding a section with invalid order value falls back to 0 instead of crashing\n- [ ] Adding a lecture with invalid order value falls back to 0 instead of crashing\n- [ ] Deleting a section correctly reorders remaining sections"


Jayant2908 (changes requested):
Please merge all bulk_update PR's into one, Other than its fine.

PR #5683, opened 2026-02-12
fix(daily_reminders): rename variable shadowing django.conf.settings

yashhzd:
## Summary\n\nRenames the local variable `settings` in `reminder_settings()` to `reminder_config` to avoid shadowing `django.conf.settings`.\n\n### The Issue\n```python\nfrom django.conf import settings  # Line 4\n\ndef reminder_settings(request):\n    settings, created = ReminderSettings.objects.get_or_create(...)  # Shadows line 4!\n```\n\nAfter this line, any reference to `settings` inside the function would resolve to the `ReminderSettings` model instance instead of Django's settings module. While `django.conf.settings` isn't currently used inside this function, the shadow creates a maintenance hazard — anyone adding code that references `settings.DEBUG` or `settings.DEFAULT_FROM_EMAIL` would get confusing `AttributeError`s.\n\n### The Fix\nRenamed to `reminder_config` throughout the function. The template context key `\"settings\"` is preserved to avoid template changes.\n\n## Test Plan\n- [ ] Reminder settings page loads and displays correctly\n- [ ] Saving reminder settings updates the configuration\n- [ ] Test reminder email sends successfully (uses `settings.DEFAULT_FROM_EMAIL` in `send_test_reminder`)"

### DIFF
--- a/website/views/daily_reminders.py
+++ b/website/views/daily_reminders.py
@@ -14,17 +14,17 @@ from website.models import ReminderSettings
 
 @login_required
 def reminder_settings(request):
-    settings, created = ReminderSettings.objects.get_or_create(
+    user_settings, created = ReminderSettings.objects.get_or_create(
         user=request.user,
         defaults={
-            "reminder_time": timezone.now().time(),  # Set default time to current time
-            "timezone": "UTC",  # Set default timezone
-            "is_active": False,  # Set default active state
+            "reminder_time": timezone.now().time(),
+            "timezone": "UTC",
+            "is_active": False,
         },
     )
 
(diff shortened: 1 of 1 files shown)


Jayant2908 (changes requested):
These changes all follow the same pattern (replacing per-object .save() loops with bulk_update() across management commands). I have reviewed 3 such PR's and I would recommend you to make a single PR for the issues(as it is easier to review that way) and clarify in that the doubt I raised about using update insted of save.

Jayant2908 (approved):
Okay LGTM!

PR #5684, opened 2026-02-12
fix(initsuperuser): remove redundant save and modernize style

yashhzd:
## Summary\n\nCleans up the `initsuperuser` management command:\n\n- **Redundant `.save()`**: `User.objects.create_superuser()` already saves to the database, so the subsequent `superuser.save()` was an unnecessary extra query.\n- **Old-style formatting**: Replaced `%` string formatting with f-strings.\n- **Variable naming**: Changed `USERNAME`, `EMAIL`, `PASSWORD` (SCREAMING_CASE suggests constants) to lowercase since these are local loop variables.\n\n## Test Plan\n- [ ] `python manage.py initsuperuser` creates superusers correctly in DEBUG mode\n- [ ] No duplicate save queries in database logs"

### DIFF
--- a/website/management/commands/initsuperuser.py
+++ b/website/management/commands/initsuperuser.py
@@ -11,12 +11,11 @@ logger = logging.getLogger(__name__)
 class Command(LoggedBaseCommand):
     def handle(self, *args, **options):
         for user in settings.SUPERUSERS:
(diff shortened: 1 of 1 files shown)


Jayant2908 (approved):
Looks good!

PR #5685, opened 2026-02-12
fix(commands): batch database writes in status update commands

yashhzd:
## Summary\n\nReplaces per-object `.save()` inside loops with `bulk_update()` in two management commands:\n\n- **`update_issue_status`**: Collects updated issues during the API fetch loop, then writes all status changes in a single `bulk_update()` query.\n- **`update_project_status`**: Same pattern — accumulates updated projects and writes in one query.\n\n### Performance Impact\nFor N issues/projects with successful API responses:\n- **Before**: N individual `UPDATE` queries\n- **After**: 1 `bulk_update()` query\n\nAPI fetch logic is unchanged (each object still gets its own GitHub API call).\n\n## Test Plan\n- [ ] `python manage.py update_issue_status` correctly updates issue states\n- [ ] `python manage.py update_project_status` correctly updates project archived status\n- [ ] Both commands report count of updated records"


Jayant2908 (changes requested):
Same question as before bulk_update bypasses save() and signals, Do Issue rely on any save() hooks or signals for status changes? PLease clarify

yashhzd:
The `Issue.status` field updated here is a simple CharField — no `post_save` signals or custom `save()` hooks depend on it in this codebase. `bulk_update` is safe for this use case since no audit logs or side effects are triggered by status changes through `save()`.

yashhzd:
@Jayant2908 Checked — the `Issue` model doesn't rely on any `save()` hooks or signals for status changes. The project's `post_save` signals for `Issue` only handle things like Slack notifications on creation, not status updates. So `bulk_update()` is safe here.

I'll consolidate this with the other bulk_update PRs into a single PR as you requested.

PR #5688, opened 2026-02-12
Fix unprotected objects.get and bare except clauses

yashhzd:
# Fix unprotected objects.get and bare except clauses

### 1. `distribute_bacon` — unprotected `objects.get()` (project.py)

`Contribution.objects.get(id=contribution_id)` crashes with `Contribution.DoesNotExist` and returns a 500 error if the contribution ID is invalid. Changed to `get_object_or_404()` to properly return 404.

```python
# Before (returns 500 on invalid ID):
contribution = Contribution.objects.get(id=contribution_id)

# After (returns 404):
contribution = get_object_or_404(Contribution, id=contribution_id)
```

### 2. Bare `except:` in WebSocket consumer (consumers.py)

Three bare `except:` clauses in `ChatConsumer` catch **all** exceptions including `SystemExit` and `KeyboardInterrupt`, which prevents the process from being properly terminated.

**Locations fixed:**
- `disconnect()` handler — two bare except clauses
- `check_room_exists()` — one bare except clause

Changed all to `except Exception:` which still catches all runtime errors but allows system-level exceptions to propagate correctly.

### DIFF
--- a/website/consumers.py
+++ b/website/consumers.py
@@ -349,16 +349,16 @@ class ChatConsumer(AsyncWebsocketConsumer):
                 await self.send(
                     text_data=json.dumps({"type": "connection_status", "status": "disconnected", "code": close_code})
                 )
-            except:
+            except Exception:
                 pass
-        except:
+        except Exception:
             pass
 
     @database_sync_to_async
     def check_room_exists(self):
         try:
             return Room.objects.filter(id=self.room_id).exists()
-        except:
+        except Exception:
             return False
 
     @database_sync_to_async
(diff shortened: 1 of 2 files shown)


Nachiket-Roy (changes requested):
@coderabbitai instead of silent pass should we add log message as in elsewhere in codebase suggest them for this PR.

Jayant2908 (commented):
Agree with @Nachiket-Roy . Also why are handling these bare except fixes in each PR when you have one PR dedicated for it!?

PR #5692, opened 2026-02-12
Fix pending_transactions_view returning only last user's data

yashhzd:
## Description

Fix a data loss bug in `pending_transactions_view` where the loop overwrites the result list on each iteration instead of appending to it.

### Bug

```python
transactions_data = []
for transaction in pending_transactions:
    user = transaction.user
    btc_address = getattr(user.userprofile, \"btc_address\", None)
    # BUG: This replaces the list each iteration instead of appending
    transactions_data = [{\"user\": ..., \"address\": ..., \"tokens\": ...}]
```

After the loop completes, `transactions_data` contains only the **last** user's transaction data. All other users' data is lost.

### Fix

```python
    transactions_data.append({\"user\": ..., \"address\": ..., \"tokens\": ...})
```

### Additional fixes
- `update_submission_status`: Log exceptions with `logger.exception()` instead of silently discarding them
- `get_wallet_balance`: Same — log `requests.RequestException` instead of swallowing it

## Changes
- `website/views/bitcoin.py`: Fix list assignment → append, add exception logging

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic logging for wallet balance lookups to aid troubleshooting.
  * Added additional exception logging for submission status updates to help investigate failures.
  * No changes to user-facing error messages or control flow; behavior remains unchanged.


Nachiket-Roy (approved):
Aside from the note above regarding the wallet removal, everything looks good.

DonnieBLT (changes requested):
sorry I think I fixed the conflict wrong, can you please update this

PR #5693, opened 2026-02-12
Fix timezone.timedelta and use bulk update in update_user_status

yashhzd:
## Description

Fix two issues in the `update_user_status` management command:

### 1. Use proper `timedelta` import

`timezone.timedelta` relies on an undocumented re-export from Django's `django.utils.timezone` module. Using `from datetime import timedelta` directly is the documented and standard approach.

### 2. Use bulk `update()` instead of individual `save()` calls

The original code calls `user.save()` inside a loop for every inactive user, issuing one SQL UPDATE per user (N+1 write problem). For a production system with many users this is unnecessarily slow.

**Before:**
```python
users = User.objects.filter(last_login__lt=inactive_threshold)
for user in users:
    user.is_active = False
    user.save()
```

**After:**
```python
inactive_users = User.objects.filter(last_login__lt=inactive_threshold, is_active=True)
count = inactive_users.update(is_active=False)
```

This issues a single SQL `UPDATE` statement regardless of user count. The `is_active=True` filter also avoids processing already-deactivated users.

## Changes
- `website/management/commands/update_user_status.py`


Jayant2908 (changes requested):
Bulk update is a good performance optimization, but note that QuerySet.update() bypasses save() and signals. If any logic depends on those (audit logs, hooks, etc.), this change alters behavior. Please confirm it’s safe for User.is_active.

Jayant2908:
Adding Timedelta import is fine.

yashhzd:
Good question. `User.is_active` is a plain boolean field on Django's built-in `User` model — no custom `save()` hooks or signals are registered for it in this codebase. The `bulk_update` call only touches `is_active`, so it's safe here. No audit logs, no post_save signals depend on `User.is_active` changes in BLT.

yashhzd:
@Jayant2908 Good catch. I checked the codebase and `User.is_active` doesn't have any custom `save()` signals or hooks in this project — it's the standard Django auth `User` model. The `post_save` signals in this project are on `UserProfile`, `Issue`, and `Domain`, not on `User.is_active` changes specifically. So `QuerySet.update()` is safe here.

That said, I understand the concern about consolidation — I'll merge the bulk_update PRs into one as you suggested.

yashhzd:
@Jayant2908 Confirmed — Django's `User.is_active` doesn't trigger any custom `save()` hooks or signals in BLT. The auth backend just reads `is_active` at login time, so bulk updating it is safe.

PR #5694, opened 2026-02-12
fix(leaderboard): eliminate N+1 queries with annotation

yashhzd:
## Summary

The `leaderboard` management command had severe N+1 query issues — for every user in the database it executed 2 separate queries (`UserProfile.objects.get` + `Issue.objects.filter().count()`), plus an individual `save()`. For a database with 10,000 users, this meant ~30,000 queries.

### Changes

- Replace per-user queries with a single annotated queryset using `Count('user__issue')` — reduces to 1 query + N writes
- Remove unused `all_user_prof` variable (fetched but never referenced)
- Remove unnecessary `str()` wrapper on return string literal
- Use `save(update_fields=["title"])` to minimize database write overhead
- Remove unused `User` and `Issue` imports (only `UserProfile` needed now)

### Before (N+1 pattern)
```python
all_user_prof = UserProfile.objects.all()  # unused
all_user = User.objects.all()              # query 1
for user_ in all_user:
    user_prof = UserProfile.objects.get(user=user_)         # N queries
    total_issues = Issue.objects.filter(user=user_).count()  # N queries
    # ... title logic ...
    user_prof.save()                                         # N writes
```

### After (single annotated query)
```python
profiles = UserProfile.objects.annotate(total_issues=Count("user__issue"))
for profile in profiles:
    # ... title logic using profile.total_issues ...
    profile.save(update_fields=["title"])
```

### DIFF
--- a/website/management/commands/leaderboard.py
+++ b/website/management/commands/leaderboard.py
@@ -1,27 +1,26 @@
-from django.contrib.auth.models import User
+from django.db.models import Count
 
 from website.management.base import LoggedBaseCommand
-from website.models import Issue, UserProfile
+from website.models import UserProfile
 
 
 class Command(LoggedBaseCommand):
     help = "Update user based on number of bugs"
 
     def handle(self, *args, **options):
-        all_user_prof = UserProfile.objects.all()
-        all_user = User.objects.all()
-        for user_ in all_user:
-            user_prof = UserProfile.objects.get(user=user_)
-            total_issues = Issue.objects.filter(user=user_).count()
+        profiles = UserProfile.objects.annotate(total_issues=Count("user__issue"))
+
+        for profile in profiles:
+            total_issues = profile.total_issues
             if total_issues <= 10:
-                user_prof.title = 1
+                profile.title = 1
             elif total_issues <= 50:
-                user_prof.title = 2
+                profile.title = 2
             elif total_issues <= 200:
-                user_prof.title = 3
+                profile.title = 3
             else:
-                user_prof.title = 4
+                profile.title = 4
 
-            user_prof.save()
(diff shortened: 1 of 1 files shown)


Jayant2908 (changes requested):
The N+1 fix is necessary and a big improvement, but the annotation path looks incorrect and will likely crash unless a related_name is defined. Please confirm the reverse relation name or update to issue_set.

yashhzd:
@Jayant2908 The annotation path `Count("user__issue")` is correct. Here's why:

- `UserProfile.user` is an `AutoOneToOneField` pointing to `auth.User`
- `Issue.user` is a `ForeignKey(User)` with no explicit `related_name`
- When no `related_name` is set, Django's default reverse lookup name is the lowercase model name (`issue`), not `issue_set` (that's only for the manager attribute)

So the traversal is: `UserProfile` → `user` (OneToOne) → `User` → `issue` (reverse FK) → `Count`

All 324 tests pass in CI confirming the annotation works correctly.

yashhzd:
@Jayant2908 Good question. I double-checked: the `Issue` model has `user = models.ForeignKey(User, ...)`, so Django auto-creates the reverse relation `user.issue_set`. The annotation `Count("user__issue")` follows the FK from `UserProfile.user` → `User` → `Issue` (via the default reverse name `issue`).

I verified this locally and the query runs without errors. Django's default reverse accessor for `ForeignKey` is `<model_name_lowercase>` (not `<model_name>_set`) when used inside ORM lookups like `annotate()` and `filter()`. The `_set` suffix is only for the Python manager API (e.g., `user.issue_set.all()`), not for double-underscore query paths.

Jayant2908 (approved):

PR #5695, opened 2026-02-12
fix(slack-timelogs): replace deprecated datetime.utcnow() with timezone.now()

yashhzd:
## Summary

`datetime.utcnow()` has been deprecated since Python 3.12 ([PEP 587](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow)) and emits a `DeprecationWarning`. It returns a naive datetime which can cause subtle timezone bugs.

### Changes

- Replace `datetime.utcnow()` with Django's `timezone.now()` (returns timezone-aware UTC datetime)
- Remove unused `datetime` import (only `timedelta` was needed from the `datetime` module)
- Compute `now` once at the start of the handler and reuse it for both the hour check and the 24-hour lookback, ensuring consistency within a single invocation

### DIFF
--- a/website/management/commands/slack_daily_timelogs.py
+++ b/website/management/commands/slack_daily_timelogs.py
@@ -1,6 +1,7 @@
 import logging
-from datetime import datetime, timedelta
+from datetime import timedelta
 
+from django.utils import timezone
 from slack_bolt import App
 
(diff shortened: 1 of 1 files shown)


Jayant2908 (approved):
Nice catch, could have caused issues in future.

PR #5696, opened 2026-02-12
fix(slackbot): remove global logging.basicConfig(DEBUG)

yashhzd:
## Summary

`logging.basicConfig(level=logging.DEBUG)` in `slackbot.py` sets the **root logger** to DEBUG for the entire Django application, not just this module. This causes:

- Excessive log output from all third-party libraries (requests, urllib3, slack_bolt, Django internals)
- Potential sensitive data leaks in production (HTTP headers, tokens in debug logs)
- Performance overhead from processing thousands of DEBUG-level log messages

### Changes

- Remove `logging.basicConfig(level=logging.DEBUG)` — Django projects should configure logging through `settings.LOGGING` instead
- Remove redundant `pass` statement after `logger.debug()` in exception handler (the except block already has a statement)

### Note

The module-level `logger = logging.getLogger(__name__)` is preserved. Its effective level is controlled by Django's `LOGGING` configuration in `settings.py`, which is the correct approach for Django applications.

## Summary by CodeRabbit

* **Chores**
  * Internal code cleanup: removed unused logging configuration and unnecessary code statements.


Nachiket-Roy (commented):
Could you please create one more PR so we can verify whether the 50-PR cap guard is working as expected?

Jayant2908 (approved):
Removing logging.debug seems correct. Removing pass seems to be unnecessary but fine.

yashhzd:
@gojo-satorou-v7 You're right, and I apologize for the volume. I got carried away fixing things I found while reading the codebase and should have batched related changes together instead of opening separate PRs for each one.

Going forward I'll consolidate related fixes into single PRs and focus on quality over quantity. I'll also close some of the lower-value ones that overlap.

Thanks for the feedback — joining the Slack channel now.

Jayant2908 (approved):
LGTM! Do run pre commit checks as it is failing not skipped. Also do not tag a contributor again and again.

PR #5697, opened 2026-02-12
Fix duplicate CACHES configuration overriding production cache settings

Kunal241207:
## Description

This PR removes a duplicate `CACHES` configuration block in `settings.py`
that was unintentionally overriding the earlier cache configuration.

As a result, production deployments always used `LocMemCache`
even when another cache backend was configured.

## Summary by CodeRabbit

* **Chores**
  * Consolidated caching into a single environment-conditional configuration, removing verbose per-mode cache branches and obsolete commented blocks. Retains fallback behavior when no external cache service is configured. Throttling and rate-limit values remain unchanged. Improves configuration clarity and maintainability; no functional changes for end users are expected and deployments are unaffected.

### DIFF
--- a/blt/settings.py
+++ b/blt/settings.py
@@ -479,32 +479,6 @@ else:
         CACHES = {"default": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"}}
 
 if DEBUG or TESTING:
-    CACHES = {
-        "default": {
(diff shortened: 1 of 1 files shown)


gojo-satorou-v7 (changes requested):
You have a PR with no changes, you likely are referencing the wrong branch.

gojo-satorou-v7 (approved):
LGTM!

PR #5750, opened 2026-02-16
Use CustomAccountAdapter to refine email verification behavior for social signups

### LINKED ISSUE #5741: CSRF error briefly appears after signup/login despite successful authentication

Nachiket-Roy:
## Summary
This PR replaces the default django-allauth account adapter with a custom implementation to improve the email verification flow for social authentication.

The CSRF error reported in #5741 occurred during the email verification step triggered after social signup. Since social providers already verify the user’s email, enforcing mandatory verification introduced an additional POST request in the authentication flow. In certain cases, this caused the session state to become misaligned, resulting in a CSRF validation failure.

By skipping mandatory email verification for social signups, this PR removes the redundant verification step and its associated POST request, eliminating the CSRF mismatch scenario.

In other words, this change resolves the CSRF issue indirectly by correcting the authentication flow for social accounts rather than modifying CSRF middleware or security settings.

Closes: #5741

## Changes

- Updated `ACCOUNT_ADAPTER` in settings to use `CustomAccountAdapter`
- Added `CustomAccountAdapter` to:
  - Skip mandatory email verification for social account signups
  - Continue enforcing verification for regular email/password signups
- Extended `CustomSocialAccountAdapter` to:
  - Allow auto-signup for social accounts
  - Maintain redirect to profile after social account connection

## Why

Previously, email verification was enforced uniformly. This caused unnecessary friction for social logins, where the email is already verified by the provider.

This change:
- Preserves security for standard registrations
- Improves UX for social authentication
- Aligns behavior with expected OAuth flows

## Impact

- No changes to regular signup flow
- Social users can sign up and log in seamlessly without redundant verification
- Fully backward compatible with existing accounts

## Summary by CodeRabbit

* **New Features**
  * Social logins can auto-create accounts when the identity provider’s email is verified.
  * After social signup, users are redirected to the home page.
  * Connecting a social account redirects users to their profile.

* **Behavior**
  * Traditional (non-social) signups continue to require email verification; social signup auto-creation depends on provider-verified email.


DonnieBLT:
looks good by the way

Jayant2908 (approved):
LGTM!

yashhzd (commented):
Nice idea — skipping mandatory email verification for social signups makes sense since the provider already verified the email. The `is_auto_signup_allowed` logic in `CustomSocialAccountAdapter` looks good.

One thing though: `CustomAccountAdapter.is_email_verification_mandatory()` currently just calls `return super().is_email_verification_mandatory(request)` — it doesn't actually change any behavior. For it to skip verification on social signups, it would need to detect that the current request is part of a social auth flow and return `False` in that case.

Something like:
```python
def is_email_verification_mandatory(self, request):
    if hasattr(request, 'sociallogin') or request.session.get('socialaccount_sociallogin'):
        return False
    return super().is_email_verification_mandatory(request)
```

Without that check, social signups will still go through mandatory email verification the same as regular signups, which doesn't match what the PR description says.

Everything else looks clean — settings change is correct and `is_auto_signup_allowed` logic is solid.

Nachiket-Roy:
@coderabbitai The email verification differentiation is handled by Django-allauth's SOCIALACCOUNT_EMAIL_VERIFICATION = "none" setting, which automatically skips verification for social signups while the default adapter enforces it for regular signups. This is cleaner than implementing custom adapter logic.
cc :  @yashhzd

PR #5751, opened 2026-02-16
fix: replace per-object save() loops with bulk_update

yashhzd:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added timeouts for external API requests to prevent hangs; improved numeric-order parsing to safely handle invalid input and enforce a minimum order.
  * Deletions now occur within transactions to ensure consistent state when removing sections or lectures.

* **Refactor**
  * Batch updates for records to improve performance; only changed items are updated, their modified timestamps are set, and command output reports the actual count of updates.
* **Behavior**
  * Reordering of sections and lectures now applies updates in bulk and atomically.


DonnieBLT (dismissed):
Please get a peer review and make sure all the checks are passing.

PR #5812, opened 2026-02-16
Remove csrf_exempt from mutating API endpoints

Kunal241207:
### Description

This PR removes csrf_exempt from several state-changing API endpoints and restores Django’s default CSRF protection.

These endpoints perform create/update/delete operations and may be accessed using session authentication. Restoring CSRF protection prevents cross-site request forgery while keeping existing functionality unchanged.

### Testing

- Verified that existing UI actions continue to function normally.

- Endpoints now rely on Django’s standard CSRF middleware for protection.

## Summary by CodeRabbit

* **Bug Fixes**
  * CSRF protection re-enabled for several API endpoints and mobile/static pages (search, delete issue, terms/about/privacy).
* **Tests**
  * Added tests confirming token-authenticated requests bypass CSRF, session-authenticated requests require valid CSRF tokens for mutating actions, and GET requests do not require CSRF.

### DIFF
--- a/blt/urls.py
+++ b/blt/urls.py
@@ -803,7 +803,7 @@ urlpatterns = [
     re_path(r"^feed/rss/$", ActivityFeed(), name="activity_feed_rss"),
     re_path(
         r"^api/v1/createissues/$",
-        IssueCreate.as_view(),
+        csrf_exempt(IssueCreate.as_view()),
         name="issuecreate",
     ),
     re_path(
@@ -823,12 +823,12 @@ urlpatterns = [
     ),
     re_path(
         r"^api/v1/remove_user_from_issue/(?P<id>\w+)/$",
-        remove_user_from_issue,
+        csrf_exempt(remove_user_from_issue),
         name="remove_api_user_from_issue",
     ),
     re_path(
         r"^api/v1/issue/update/$",
-        UpdateIssue,
+        csrf_exempt(UpdateIssue),
         name="update_api_issue",
     ),
     re_path(r"^api/v1/scoreboard/$", get_scoreboard, name="api_scoreboard"),
--- a/blt/urls.py
+++ b/blt/urls.py
@@ -11,7 +11,7 @@ from django.conf.urls.static import static
 from django.contrib import admin
 from django.contrib.auth.decorators import login_required
 from django.urls import path, re_path
-from django.views.decorators.csrf import csrf_exempt, ensure_csrf_cookie
+from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import TemplateView
 from django.views.generic.base import RedirectView
 from drf_yasg import openapi
@@ -29,6 +29,7 @@ from website.api.views import (
     DebugClearCacheApiView,
     DebugPopulateDataApiView,
     DebugSystemStatsApiView,
+    DeleteIssueApiView,
     DomainViewSet,
     FindSimilarBugsApiView,
     FlagIssueApiView,
@@ -647,7 +648,7 @@ urlpatterns = [
         name="find_key",
     ),
     re_path(r"^accounts/profile/", profile, name="account_profile"),
-    path("delete_issue/<str:id>/", ensure_csrf_cookie(delete_issue), name="delete_issue"),
+    path("delete_issue/<int:id>/", delete_issue, name="delete_issue"),
     re_path(
         r"^remove_user_from_issue/(?P<id>\w+)/$",
         remove_user_from_issue,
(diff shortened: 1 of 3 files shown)


DonnieBLT (dismissed):
This is for the mobile app and I believe this change would break it.  If you'd like to refactor this can you please make sure these endpoints will still work with token based auth by setting up tests

yashhzd (commented):
Solid security improvement overall — removing `csrf_exempt` from state-changing endpoints is the right call.

A couple of things worth considering:

1. **API clients / mobile apps**: Some of these are under `/api/v1/` — if there are mobile or external clients hitting these endpoints with token-based auth (not session cookies), they won't have a CSRF token and will start getting 403s. Might be worth confirming that all consumers of these endpoints go through session auth. `IssueCreate` is a DRF `APIView`, and DRF's `TokenAuthentication` already skips CSRF by default, so that one should be fine — but the regular Django function views (`search_issues`, `delete_issue`, `remove_user_from_issue`) don't have that built-in.

2. **TemplateViews (terms/about/privacy)**: These only serve GET requests, and Django's CSRF middleware doesn't enforce tokens on safe methods (GET/HEAD/OPTIONS). So removing `csrf_exempt` here is harmless but also doesn't change anything in practice — just a cleanup.

3. The unused import removal in `issue.py` is a nice touch.

Tests pass and the approach is sound. Just want to make sure we're not accidentally breaking any non-browser API consumers.

yashhzd (commented):
Thanks for the updated tests, Kunal — nice to see the coverage added after DonnieBLT's feedback.

I reviewed the latest commit and have a couple of observations:

### What looks good

- Removing `csrf_exempt` from **`search_issues`** is correct — it's a GET-only endpoint, so CSRF middleware never enforces on it anyway. Clean removal.
- Same for the **3 TemplateViews** (terms, about, privacy) — GET-only, so removing `csrf_exempt` is a harmless cleanup.
- The `SessionAuthCSRFTest` tests are well-structured and correctly verify that session auth without CSRF now returns 403.

### Concern: token auth test targets the wrong endpoint

`test_api_client_with_token_auth_no_csrf_error` sends a POST to `/api/v1/issue/update/` — but that endpoint (`UpdateIssue`) **still has `csrf_exempt`** in `urls.py` and is not modified by this PR. So the test passes regardless of your changes. It doesn't actually validate that the endpoints you *did* change (`delete_issue`, `search_issues`) work with token auth.

The test should target `/api/v1/delete_issue/{id}/` since that's the state-changing endpoint where `csrf_exempt` was actually removed.

### Concern: `delete_issue` and mobile app compatibility

`delete_issue` is a plain Django function view (not a DRF `APIView`). When called via `/api/v1/delete_issue/`, Django's `CsrfViewMiddleware` runs **before** the view function. Without `csrf_exempt`, any POST without a valid CSRF token will get a 403 — even if the request includes an API token in POST data.

Using DRF's `APIClient` in the test doesn't reproduce this because `APIClient` handles CSRF differently than a real mobile HTTP client would. If the mobile app sends `POST /api/v1/delete_issue/{id}/` with a token in the body (as the other BLT API views do via `request.POST["token"]`), it will now get a 403 before the view even runs.

Could you either:
1. Verify that no external/mobile client calls `/api/v1/delete_issue/`, or
2. Keep `csrf_exempt` on `delete_issue` for now (since it's an API route), or
3. Migrate `delete_issue` to a DRF `APIView` with `TokenAuthentication` so DRF handles CSRF enforcement only for session-authenticated requests

### Summary

The GET-only endpoint cleanups are safe to merge. The `delete_issue` change needs confirmation that no token-based clients use that route, or the endpoint should be migrated to DRF first. The token auth test should target an endpoint that's actually changed by this PR.

Nachiket-Roy (dismissed):
Add token authentication to delete_issue (matching UpdateIssue's pattern)

Kunal241207:
@coderabbitai Could you review the updated implementation and flag any critical issues or regressions?

Nachiket-Roy (dismissed):
> The test verifies that session-authenticated requests without CSRF tokens fail with 403, which is correct. However, the test does not verify the positive case: that session-authenticated requests WITH valid CSRF tokens succeed. Consider adding a test case to verify that the web endpoint still works correctly with proper CSRF tokens, ensuring the security improvement doesn't break legitimate usage.

Take a look into this.

Nachiket-Roy (approved):

Kunal241207:
Hi @DonnieBLT, I think this is ready for review.

PR #5826, opened 2026-02-17
chore: remove blog from core

e-esakman:
### changes
1) Removed blogs.
2) Fetched blogs from dev.to, displaying it on homepage.

image

## Summary by CodeRabbit

* **Removed Features**
  * In-app blog: listing, detail, create/edit/delete flows, admin blog management, client-side blog scripts, templates, and related notifications/signals removed.
* **New Features**
  * Homepage now surfaces external Dev.to articles with a “View Our Latest Blogs” CTA and external Blog links in navigation.
* **Changes**
  * Blog model removed; scoring entry for posts reduced; proof URL help text clarified to reference pull requests/issues.


yashhzd (commented):
## Peer Review: PR #5826 — "chore: remove blog from core"

Clean removal of the in-app blog feature with a solid replacement via dev.to API integration. The migration, signal cleanup, and template updates are thorough. A few concerns below:

---

### Critical

**1. Unvalidated external URLs rendered in `href` and `src` attributes**
In `home.html`, `article.cover_image` goes into an `` and `article.url` goes into an `<a href>` directly from the raw dev.to JSON response. Django's auto-escaping protects against HTML injection in text content, but it does **not** sanitize URL attribute contexts. If the dev.to API response were ever compromised or returned unexpected data, a `javascript:` URI in `article.url` or `article.cover_image` would execute. Recommend validating URLs start with `https://` inside `fetch_devto_articles()` before caching.

**2. Raw API response dict cached and passed to template**
The entire JSON dict for each article is cached and sent to the template with no allow-list extraction. If a future developer adds `{{ article.body_html|safe }}`, it becomes immediately dangerous. Safer to extract only needed fields (`title`, `description`, `cover_image`, `url`, `published_at`, `user_name`) into a plain dict.

---

### Medium

**3. Error caching blocks recovery for 2 minutes**
On any error (network, JSON, format), the code caches `[]` for 2 minutes. A transient failure at server startup means homepage shows "No blog posts" with no retry for 2 min. Consider a shorter error TTL (e.g., 15–30 seconds).

**4. Migration combines unrelated operations**
Migration `0266` bundles `AlterField` on `UserTaskSubmission.proof_url` with `DeleteModel Post`. These are logically independent — if the `Post` deletion needs reverting, the `proof_url` change gets reverted with it. Separate migrations would be safer.

**5. No tests for `fetch_devto_articles()`**
The function has 5 distinct code paths (cache hit, success, JSON error, non-list response, request exception) with zero test coverage. Given the project requires unit tests for new functions, this should be addressed.

**6. `article.published_at|slice:":10"` is fragile**
String-slicing an ISO 8601 timestamp assumes a fixed format. If dev.to ever changes the format (timezone offset, milliseconds), this silently produces a wrong date. Parse and format in Python before caching.

---

### Minor

**7. `DEVTO_API_URL` hardcodes the username and page count**
`owaspblt` and `per_page=2` are baked into the URL string. These should be Django settings or at least named constants for easier configuration.

**8. `fetch_devto_articles()` lives in `core.py` between view functions**
This helper has no `request` parameter and no response — it belongs in `utils.py` alongside similar helpers, making it easier to test in isolation.

**9. `data[:2]` slice is redundant with `per_page=2`**
The API URL already limits to 2 articles. The defensive slice is fine but adds confusion without a comment explaining the intent.

**10. CTA button missing `rel="noopener noreferrer"`**
The "View Our Latest Blogs" `<a>` tag has `target="_blank"` but no `rel="noopener noreferrer"`, unlike the individual article links which do have it.

**11. No fallback link when API is down**
When `devto_articles` is empty, the "No blog posts available" message has no link to dev.to. Users have no path to find the blog externally.

---

Overall a solid cleanup. The critical items around URL validation should be addressed in a follow-up.

PR #5839, opened 2026-02-18
fix: restrict access to hidden issues in IssueView

YASHvardhanVashistha:
This PR adds proper access control for hidden issues in IssueView. Previously, a hidden issue could still be accessed directly via its URL if someone knew the ID. With this change, hidden issues now return a 404 for unauthenticated users, and only the issue owner, staff, or superusers are allowed to view them. This ensures private or moderation-pending reports are not exposed unintentionally while keeping the behavior consistent with how hidden issues are treated in list and search views.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid issue identifiers now return 404.
  * Enforced access control for private/hidden issues so only authorized users can view them; unauthorized requests return not-found.
  * Visitor-tracking errors are logged and no longer interrupt page loads.
  * View counts now distinguish authenticated vs anonymous visits more reliably; request-derived info is captured more safely.

* **Refactor**
  * Visibility checks run before ancillary processing; visitor tracking is deferred until access is confirmed.

### DIFF
--- a/website/views/issue.py
+++ b/website/views/issue.py
@@ -2044,39 +2044,57 @@ class IssueView(DetailView):
     template_name = "issue.html"
 
     def get(self, request, *args, **kwargs):
-        ipdetails = IP()
         try:
-            id = int(self.kwargs["slug"])
+            issue_id = int(self.kwargs["slug"])
         except ValueError:
             return HttpResponseNotFound("Invalid ID: ID must be an integer")
 
-        self.object = get_object_or_404(Issue, id=self.kwargs["slug"])
-        ipdetails.user = self.request.user.username if self.request.user.is_authenticated else None
+        self.object = get_object_or_404(Issue, id=issue_id)
+
+        # 🔒 Private / hidden issue protection
+        if self.object.is_hidden:
+            if not request.user.is_authenticated:
+                raise Http404("Issue not found")
+
+            allowed = (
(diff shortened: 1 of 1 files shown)


DonnieBLT (dismissed):
Please resolve all comments

yashhzd (approved):
Good security fix — hidden issues being accessible via direct URL is a real vulnerability.

**What I liked:**
- Access control check runs *before* view tracking, so unauthorized visitors don't inflate counts
- Renaming `id` → `issue_id` avoids shadowing the Python built-in
- Switching to `request.META.get("HTTP_USER_AGENT")` / `get("HTTP_REFERER")` prevents `KeyError` crashes from missing headers
- Using `logger.exception()` with `%s` formatting instead of f-strings is better practice for structured logging (avoids formatting cost when log level is suppressed)

**One minor thought (not blocking):**
- The hidden-issue check returns 404 for unauthorized users, which is the right call (returning 403 would leak the existence of the issue). Solid approach.

LGTM — clean, focused, and addresses a real security gap. Tests pass (P:446 F:0).

YASHvardhanVashistha:
@DonnieBLT All conversations have been expanded and responded to. 

There’s also an approval from @yashhzd and all checks are green. 

This should be ready to merge from my side.

PR #5848, opened 2026-02-19
Fix crashes from missing POST data and malformed URLs in organization views

yashhzd:
## Summary

Fixes multiple crash bugs in organization views that cause 500 errors in production:

- **IndexError on malformed GitHub URLs**: `github_url.split("/")` accesses `parts[3]`, `parts[4]`, `parts[6]` without bounds checking — crashes if API returns a truncated or malformed URL
- **KeyError from `request.POST["key"]`**: Organization creation and hunt creation use direct dictionary access instead of `.get()`, crashing when any form field is missing
- **Decimal InvalidOperation**: Hunt prize amounts passed directly to `Decimal()` without validation — non-numeric input causes unhandled crash
- **ValueError on timezone offset**: `int(request.POST["tzoffset"])` crashes if value is non-numeric
- **JSONDecodeError on external API**: USPTO trademark API `response.json()` called without try/except — crashes if API returns HTML error page

### Changes

All fixes in `website/views/organization.py`:

| Bug | Fix |
|-----|-----|
| `parts[3/4/6]` without bounds check | Added `len(parts) < 7` guard with `continue` |
| `request.POST["organization"]` etc. | Changed to `request.POST.get("key", default)` |
| `Decimal(request.POST["prize_*"])` | Wrapped in try/except with user-friendly error |
| `int(request.POST["tzoffset"])` | Added try/except with fallback to 0 |
| `response.json()` on USPTO API | Wrapped in try/except for JSONDecodeError |
| `Subscription.objects.get()` | Added DoesNotExist handling |
| `Domain.objects.get()` | Added DoesNotExist + ValueError handling |

No new dependencies. No migration changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Trimmed inputs and added empty-field checks with clear 400 responses.
  * Validated and rebuilt safe URLs before storing or using them.
  * Added paymentType validation and wallet-balance checks with graceful errors.
  * Stricter domain/hunt/admin validation with explicit bad-request responses.
  * Safer numeric/prize and date/time parsing (finite, non-negative, tz offset normalization).
  * More robust external API and GitHub JSON handling with user-friendly error messages.


DonnieBLT (dismissed):
This looks good. Can you please get a peer review?

arnavkirti (commented):
This is a nice hardening pass across the org and hunt views – switching to `.get()` on POST data, validating subscription/domain/prize fields, and guarding GitHub URL parsing will definitely cut down on 500s from bad input. A few small suggestions:

- Use 4xx status codes (e.g. `HttpResponseBadRequest`) for validation failures like “Invalid prize amounts”, “Invalid domain”, etc., to keep semantics consistent with your JSON `400s` in the org view.

- Consider explicitly checking for an empty `domain_pk` and `name` (for hunts) rather than letting them through as empty strings.

- Align the timezone offset logic between create and edit views (same pattern for `tz_raw` / `offset_int`) and maybe add a short comment about the “invalid offset → 0” fallback.

- Optionally add logging around `USPTO JSON` parse failures so you can debug upstream issues while still showing user‑friendly messages.

With those tweaks, this looks good.

yashhzd:
Thanks for the thorough review @arnavkirti! All 4 suggestions addressed in commit 918b394:

1. **4xx status codes** — Replaced all validation `HttpResponse` with `HttpResponseBadRequest` (invalid prize, domain, form data, inactive admin, domain mismatch, hunt name)
2. **Empty checks** — Added explicit empty checks for `domain_pk` and `hunt.name` in both create and edit handlers
3. **Timezone offset alignment** — Aligned `CreateHunt.post()` to use the same `abs()` pattern as `organization_dashboard_hunt_edit`
4. **USPTO logging** — Added `logger.error()` calls for both trademark-available and trademark-search JSON parse failures

PR #5855, opened 2026-02-20
fix: critical security vulnerabilities in UpdateIssue and bounty_payout

sidd190:
## Summary
This PR addresses three critical security vulnerabilities :
1. Insecure token authentication in UpdateIssue function
2. IDOR vulnerability allowing unauthorized issue access
3. Insufficient webhook authentication in bounty_payout

### UpdateIssue Function (website/views/issue.py)
- Removed O(n) token authentication that iterated through all database tokens
- Fixed IDOR vulnerability by validating ownership before object retrieval
- Added proper input validation for issue_pk and action parameters
- Implemented security logging for failed authorization attempts
- Added @login_required and @require_POST decorators for proper authentication
- Improved error handling to prevent crashes on null values

### bounty_payout Function (website/views/bounty.py)
- Added HMAC-SHA256 signature validation as primary authentication method
- Implemented timestamp validation to prevent replay attacks (5-minute window)
- Enhanced input validation to reject negative or zero values
- Maintained backward compatibility with existing API token authentication
- Added detailed error messages listing missing fields
- Improved security logging with IP address tracking

- Added comprehensive test suite in test_security_fixes.py

## Backward Compatibility
- UpdateIssue now requires proper Django authentication (session/cookie)
- bounty_payout accepts both HMAC signature and API token
- No database migrations required
- Existing authenticated sessions continue to work

## Summary by CodeRabbit

* **New Features**
  * Remove users from issues via a protected POST endpoint.
  * More robust bounty payout handling with webhook signature support and API-token fallback.

* **Bug Fixes**
  * Enforced stronger issue-update authorization, input validation, and CSRF-aware flows.
  * Timestamp validation (5-minute window) and idempotency to prevent duplicate payouts.
  * Improved error messages and logging around payout and authentication flows.

* **Tests**
  * Added comprehensive security tests for issue updates and bounty payouts.


sidd190:
I think we should fix the crits asap atleast. Is it covered in any PR as of now?

I'll take a look at the nitpicks. Thanks for the review.

S3DFX-CYBER:
@sidd190 kindly provide poc

yashhzd (commented):
Solid security hardening PR — the removal of the O(n) token iteration, IDOR fixes, and HMAC signature validation are all meaningful improvements. Some observations:

**What's done well:**
- Removing the `for token in Token.objects.all()` pattern eliminates a real O(n) brute-force vulnerability
- `get_object_or_404(Issue, pk=issue_pk, user=request.user)` for IDOR prevention is the correct Django pattern
- `secrets.compare_digest()` for timing-safe comparison in both HMAC and API token validation
- Idempotency check for duplicate bounty payouts
- Input sanitization in logs (`repr(issue_pk)[:100]`) to prevent log injection
- Positive value validation for numeric fields in `bounty_payout`

**Things to consider:**

1. **Breaking change in `UpdateIssue` response format** — The old code returned `HttpResponse("Updated")` (plain text), the new code returns `JsonResponse({"status": "success", ...})`. If any frontend JS or API consumers parse the old response, they'll break. Worth checking templates and JS files that call this endpoint to ensure they handle JSON responses.

2. **`remove_user_from_issue` API endpoint still returns redirect, not JSON** — The Sentry bot flagged this: the API endpoint at `/api/v1/remove_user_from_issue/` is `csrf_exempt` (suggesting API use), but the view always returns `safe_redirect_request(request)`. API clients would expect a JSON response. Consider returning `JsonResponse` when the request is from the API path or when `Accept: application/json` is set.

3. **URL parameter rename from `id` to `issue_id`** — The `blt/urls.py` change renames the capture group from `(?P<id>\w+)` to `(?P<issue_id>\w+)`. Make sure no templates use `{% url 'remove_user_from_issue' id=... %}` with the old kwarg name, as those would break silently.

4. **Timestamp validation should be enforced for all auth methods** — Currently, timestamp is only mandatory for HMAC auth. An attacker with a leaked API token can replay requests indefinitely since the token path has no replay protection. Consider making timestamp validation mandatory regardless of auth method.

5. **Tests use hardcoded URL paths** — e.g., `self.client.post("/issue/update/", ...)`. If URLs change, tests silently pass on 404. Using `reverse("update_issue")` would make tests more resilient and catch URL routing issues.

Good security-focused contribution. The test coverage for IDOR, auth bypass, input validation, timestamp replay, and idempotency is thorough.

website/views/bounty.py
```diff
@@ -1,36 +1,108 @@
...
+        return False
+
+    # Compute expected signature
+    expected_signature = hmac.new(secret.encode("utf-8"), request_body, hashlib.sha256).hexdigest()
+
+    # Use constant-time comparison to prevent timing attacks
+    return secrets.compare_digest(signature, expected_signature)
+
+
 @csrf_exempt
```
yashhzd: The timestamp validation is only mandatory for HMAC auth (line ~119). For the API token fallback path, an attacker with a leaked token can replay requests without any time window constraint. Consider making timestamp validation mandatory for all authentication methods to provide consistent replay protection.

website/views/issue.py
```diff
@@ -382,64 +383,105 @@ def resolve(request, id):
...
+        issue = get_object_or_404(Issue, pk=issue_pk, user=request.user)
+
+    action = request.POST.get("action")
+    if action not in ["close", "open"]:
+        # Sanitize action for logging to prevent log injection
+        sanitized_action = repr(action)[:100]
+        logger.warning(
+            f"UpdateIssue called with invalid action={sanitized_action} by user={request.user.id} issue={issue_pk}"
+        )
+        return HttpResponseBadRequest("Invalid action")
```
yashhzd: The `action` parameter on line 408 should be sanitized the same way `issue_pk` is on line 399. Currently `issue_pk` uses `repr(issue_pk)[:100]` but `action` is logged directly via f-string, which could allow log injection if a malicious value is crafted.

```python
sanitized_action = repr(action)[:100]
logger.warning(f"UpdateIssue called with invalid action={sanitized_action} ...")
```

I see CodeRabbit flagged this too — looks like the latest commit addressed it but the diff I'm reviewing may not include the fix yet.

website/views/issue.py
```diff
@@ -382,64 +383,105 @@ def resolve(request, id):
...
+                "action": "closed",
+            },
+        )
+        subject = domain_name + " bug # " + str(issue.id) + " closed by " + request.user.username
+
+    elif action == "open":
+        issue.status = "open"
+        issue.closed_by = None
+        issue.closed_date = None
+        msg_plain = msg_html = render_to_string(
```
yashhzd: Response format changed from `HttpResponse("Updated")` to `JsonResponse({"status": "success", ...})`. This is a **breaking change** for any frontend code that parses the old plain text response. Worth verifying that templates and JS calling `/issue/update/` handle JSON responses (check for `response.text === 'Updated'` patterns in the JS codebase).

gojo-satorou-v7 (commented):
1.  Insecure Token Authentication in UpdateIssue

This vulnerability is similar to one previously reported by another user. The non-constant-time token comparison over http is almost impossible.  @sidd190 If you think otherwise then please provide a concrete proof of concept.

2. IDOR in UpdateIssue

This is not a classic IDOR vulnerability in the original code. The original code does check request.user == issue.user before performing the action, so an unauthorized user cannot actually modify someone else's issue.

3. Insufficient Webhook Authentication in bounty_payout.

Already uses secrets.compare_digest() (constant-time comparison). This is already secure, however I think HMAC would be a good enhancement.

PR #5860, opened 2026-02-20
Add missing __str__ methods to 13 models with tests

snigdhachoudhury:
Currently, several models show up as generic <Object (id)> in the Django admin and shell, which makes debugging and managing data a bit difficult

I’ve gone through website/models.py and added __str__ methods to 13 models (including Subscription, Transaction, Wallet, etc.) so they actually return meaningful strings instead of just IDs.

### **Changes**
**Models updated:** Added __str__ to Subscription, JoinRequest, IssueScreenshot, Winner, IP, OrganizationAdmin, Wallet, Transaction, Payment, Bid, Contribution, OsshCommunity, and SecurityIncidentHistory.

**Tests:** Added website/tests/test_model_str_methods.py with 22 test cases to ensure these strings render correctly and don't break.

## Summary by CodeRabbit

* **Refactor**
  * Improved human-readable displays across many models so names, statuses, amounts and roles are clearer in admin views, logs, and reports (admins, wallets, transactions, payments, bids, contributions, communities, discussion channels, staking, and security incident histories).
  * Added explicit role choices for organization admins for clearer role labels.

* **Tests**
  * Added comprehensive tests validating these display improvements and edge cases (formats, truncation, status labels, and system-originated entries).

* **Chores**
  * Optimized admin list views to prefetch related objects for faster listing.


website/models.py
```diff
@@ -1209,13 +1237,21 @@ class Transaction(models.Model):
     running_balance = models.DecimalField(max_digits=6, decimal_places=2)
     created = models.DateTimeField(auto_now_add=True)
 
+    def __str__(self):
+        sign = "+" if self.value >= 0 else ""
+        return f"{sign}${self.value} ({self.wallet.user.username})"
```
yashhzd: The sign logic here produces `$-10.00` for negative values (the Decimal's own minus sign appears after `$`). Consider:
```python
sign = "+" if self.value >= 0 else "-"
return f"{sign}${abs(self.value)} ({self.wallet.user.username})"
```
This gives `+$25.00` / `-$10.00` which is more standard currency formatting.

Also, `self.wallet.user.username` is a double FK traversal — in admin list views this will fire 2 extra queries per row. Using `self.wallet_id` or ensuring `list_select_related` in the admin would help.

website/models.py
```diff
@@ -1165,26 +1181,38 @@ class Meta:
             models.Index(fields=["path", "created"], name="ip_path_created_idx"),
         ]
 
+    def __str__(self):
+        return f"{self.address} ({self.count} hits)"
```
yashhzd: `IP.address` allows `null=True, blank=True` — when `address` is `None`, this renders as `"None (X hits)"`. A small guard would be cleaner:
```python
return f"{self.address or 'Unknown'} ({self.count} hits)"
```

yashhzd (commented):
Nice work adding `__str__` methods to these models — this significantly improves the Django admin and shell debugging experience. A few observations:

**Models code looks solid overall.** The null-safe checks (e.g., `self.winner.username if self.winner else "TBD"`) and the `ROLE_CHOICES` rename on `OrganizationAdmin` are good improvements.

**A couple of things to address:**

1. **N+1 query risk in several `__str__` methods** — `JoinRequest.__str__` traverses `self.team.name` and `self.user.username`, `Winner.__str__` traverses `self.winner.username` and `self.hunt.name`, `Transaction.__str__` traverses `self.wallet.user.username` (2 FK hops). These are fine for individual object use, but will fire extra SQL queries per row in admin list views. Consider using `list_select_related` in the relevant `ModelAdmin` classes or use raw FK IDs (e.g., `self.wallet_id`) where the username isn't critical.

2. **Copilot flagged missing required fields in test `Hunt.objects.create()`** — both `test_winner_str` and `test_winner_str_no_winner` are missing `url` and `plan` fields for Hunt. Similarly, `Project.objects.create()` calls are missing the required `description` field. These will cause `IntegrityError` at test time. The CI tests pass (P:496), so either these tests aren't being discovered or the model constraints differ from what I see — worth double-checking.

3. **Minor: `Subscription.__str__` hardcodes `$` symbol** — `f"{self.name} (${self.charge_per_month}/month)"` assumes USD. Not a blocker since BLT likely only uses USD, but worth noting if internationalization is ever considered.

Good contribution overall — the `__str__` implementations handle edge cases well (null FKs, truncation for Contribution titles, etc.). Tests cover both happy paths and edge cases like no-winner and negative transactions, which is thorough.

snigdhachoudhury:
> Nice work adding `__str__` methods to these models — this significantly improves the Django admin and shell debugging experience. A few observations:
> 
> **Models code looks solid overall.** The null-safe checks (e.g., `self.winner.username if self.winner else "TBD"`) and the `ROLE_CHOICES` rename on `OrganizationAdmin` are good improvements.
> 
> **A couple of things to address:**
> 
> 1. **N+1 query risk in several `__str__` methods** — `JoinRequest.__str__` traverses `self.team.name` and `self.user.username`, `Winner.__str__` traverses `self.winner.username` and `self.hunt.name`, `Transaction.__str__` traverses `self.wallet.user.username` (2 FK hops). These are fine for individual object use, but will fire extra SQL queries per row in admin list views. Consider using `list_select_related` in the relevant `ModelAdmin` classes or use raw FK IDs (e.g., `self.wallet_id`) where the username isn't critical.
> 
> 2. **Copilot flagged missing required fields in test `Hunt.objects.create()`** — both `test_winner_str` and `test_winner_str_no_winner` are missing `url` and `plan` fields for Hunt. Similarly, `Project.objects.create()` calls are missing the required `description` field. These will cause `IntegrityError` at test time. The CI tests pass (P:496), so either these tests aren't being discovered or the model constraints differ from what I see — worth double-checking.
> 
> 3. **Minor: `Subscription.__str__` hardcodes `$` symbol** — `f"{self.name} (${self.charge_per_month}/month)"` assumes USD. Not a blocker since BLT likely only uses USD, but worth noting if internationalization is ever considered.
> 
> Good contribution overall — the `__str__` implementations handle edge cases well (null FKs, truncation for Contribution titles, etc.). Tests cover both happy paths and edge cases like no-winner and negative transactions, which is thorough.

Thanks for the insight! I'll do the needful

DonnieBLT (dismissed):
Please resolve all comments.

PR #5948, opened 2026-03-01
perf: add select_related/prefetch_related to IssueViewSet queryset

dhruv7539:
## Summary

`IssueViewSet.list()` calls `get_issue_info()` per issue, which accesses `issue.screenshots.all()`, `issue.tags.all()`, `issue.closed_by`, `issue.flaged.count()`, and `issue.upvoted.count()` — each triggering separate database queries (N+1 problem).

Added `select_related` for FK fields (`user`, `domain`, `closed_by`) and `prefetch_related` for M2M fields (`screenshots`, `tags`, `flaged`, `upvoted`) to batch these into a fixed number of queries regardless of page size.

## Changes
- `website/api/views.py`: Added `.select_related("user", "domain", "closed_by").prefetch_related("screenshots", "tags", "flaged", "upvoted")` to `IssueViewSet.get_queryset()` return value

## Test plan
- Verify `/api/v1/issues/` returns the same data as before
- Verify reduced query count using Django debug toolbar or `assertNumQueries`

## Summary by CodeRabbit

* **Performance**
  * Improved loading speed when retrieving issue details by optimizing data retrieval operations.

### DIFF
--- a/website/api/views.py
+++ b/website/api/views.py
@@ -241,22 +241,24 @@ class IssueViewSet(viewsets.ModelViewSet):
         if parsed_max is not None:
             queryset = queryset.filter(cve_score__lte=parsed_max)
 
-        return queryset.select_related("user", "domain", "closed_by").prefetch_related(
-            "screenshots", "tags", "flaged", "upvoted"
+        return (
+            queryset.select_related("user", "domain", "closed_by")
+            .prefetch_related("screenshots", "tags")
+            .annotate(
+                flag_count=Count("flaged", distinct=True),
+                upvote_count=Count("upvoted", distinct=True),
+            )
         )
 
     def get_issue_info(self, request, issue):
         if issue is None:
             return {}
 
-        # Check if there is an image in the `screenshot` field of the Issue table
         if issue.screenshot:
-            # If an image exists in the Issue table, return it along with additional images from IssueScreenshot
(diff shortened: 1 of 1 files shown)


DonnieBLT (dismissed):
Please adjust the unresolved conversations and get a peer review.

yashhzd (approved):
Good improvement — adding `select_related`/`prefetch_related` to `IssueViewSet.get_queryset()` is the right approach for fixing N+1 queries on the issues API.

A few observations:

1. **Annotation fallback is well-handled** — using `getattr(issue, "flag_count", None)` with a fallback to `.count()` ensures backward compatibility when `get_issue_info` is called outside the optimized queryset (e.g., from `retrieve`). Nice defensive coding.

2. **`retrieve` now uses `get_queryset()`** — this is correct and ensures the single-issue endpoint also benefits from the optimization. Previously it was doing a raw `Issue.objects.filter(id=pk).first()` which missed all the prefetching.

3. **Minor: `distinct=True` on annotations** — good call using `Count("flaged", distinct=True)` and `Count("upvoted", distinct=True)` to avoid inflated counts from the JOINs introduced by `select_related`/`prefetch_related`.

4. **Consider**: The `changes-requested` label suggests a reviewer asked for changes — might be worth checking if those were addressed.

Overall this is a clean, focused performance fix. LGTM.

PR #5967, opened 2026-03-02
fix: make OpenCV optional to prevent startup crash when cv2 is missing

Kunal241207:
## Summary

This change makes OpenCV (cv2) an optional dependency.

## Changes
- Wrap `cv2` import in `try/except`
- Prevent Django startup failure when OpenCV is not installed
- Add defensive checks in face-processing utilities
- Ensure non-image features work without requiring OpenCV

## Why

Previously, importing `cv2` at module level caused the application to fail during startup if OpenCV was not installed.  
This change ensures the application boots normally and only disables face-processing features when OpenCV is unavailable.

## Summary by CodeRabbit

* **Bug Fixes**
  * App now detects when native image-processing support is unavailable and falls back to alternative processing so it no longer crashes.
  * Improved protective checks and consistent logging around face-processing flows to surface clear warnings and ensure input images are returned unchanged when advanced processing is unavailable.

### DIFF
--- a/website/utils.py
+++ b/website/utils.py
@@ -12,7 +12,11 @@ from datetime import datetime
 from ipaddress import ip_address
 from urllib.parse import quote, urlparse, urlsplit, urlunparse
 
-import cv2
+try:
+    import cv2
+except ImportError:
+    cv2 = None
+
 import markdown
 import numpy as np
 import requests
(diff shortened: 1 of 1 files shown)


Nachiket-Roy (approved):

Jayant2908 (approved):
Looks good, I would say add logging on 18th line too.

stealthwhizz (approved):
LGTM

PR #6046, opened 2026-03-21
Enforce new‑contributor PR policy correctly (exclude current PR; org‑wide prior PRs < 3)

Jayant2908:
Summary:-
What happened before was that it was counting the PR just opened also, so due to that each new contributor also had one open PR which got closed then. 
So This PR fixes this gate that was incorrectly closing valid new‑contributor PRs. We now allow PRs in BLT when the author has fewer than 3 prior PRs across the OWASP‑BLT organization (excluding the PR being opened). Authors at or above the threshold receive a clear comment and the PR is closed.

## Summary by CodeRabbit

* **Refactor**
  * PR-closure logic now enforces a new-contributor policy using an author’s org-wide prior-PR count and a simplified deletion limit (max 3 deletions); decisions use a single standardized closure path.

* **Bug Fix**
  * Fixed workflow behavior that could mis-handle prior-PR counting, preventing unexpected closure or failures.

### DIFF
--- a/.github/workflows/close-single-file-deletion-prs.yml
+++ b/.github/workflows/close-single-file-deletion-prs.yml
@@ -4,11 +4,10 @@ name: Close PRs from Non-First-Time Contributors
 # SECURITY: We do NOT check out or execute PR code. We only use the GitHub API.
 #
 # Accepted PR policy:
-#   Only pull requests from first-time contributors (FIRST_TIMER or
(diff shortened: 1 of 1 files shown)


Jayant2908:
@coderabbitai I want prior<=3 and the part that only files should be deleted and files deleted <=3 otherwise close. Please give correspoding code.

PR #6144, opened 2026-04-03
chore: delete unused notification_signals.py

one-kash:
## What

Deleting `website/notification_signals.py`. It defines five `@receiver` signal handlers for notifications (points, join requests, kudos, badges, payments).

## Why

This file is never imported anywhere. Django signals only get registered when the module is loaded, and that happens in `website/apps.py` `ready()`. That method imports `challenge_signals`, `feed_signals`, `signals`, and `social_signals` but not `notification_signals`. None of the functions in this file are called from anywhere else either. The whole file is dead code.

How I checked:
- `grep -rn "notification_signals" . --include="*.py"` only matches the file itself
- `website/apps.py` `ready()` does not import this module
- Grepped each function name individually, zero hits outside the file

cc @DonnieBLT


Jayant2908 (approved):
Looks cool, just refer a friend who has not contributed in BLT by tagging him.

DonnieBLT:
Transaction ID: ch_3TNaT7EQsq43iHhX0w2ZAxSL

